PR #372, opened 2026-07-22
feat(harness): show release version at startup and in UI

### LINKED ISSUE #369: Show the release version at Harness startup and in the UI
## Agent Brief

**Category:** enhancement
**Summary:** Surface the running Ready for Agent release version in the production Harness startup output and shared UI chrome.

**Current behavior:**
The production Harness logs its listening URL when startup completes, but does not identify the version that is running. The UI also has no visible version identifier. The compiled operator binary already has a canonical version embedded from the versioned `ready-for-agent` launcher package and exposes it through `--version`; the Harness package manifests themselves retain development placeholder versions.

**Desired behavior:**
When the production Harness becomes ready, its startup message should include the running Ready for Agent version alongside the listening URL. The same version should be visible in the UI's shared application chrome so it can be found from any page without opening a dialog.

Both surfaces must use the same build-time release version as the operator binary's `--version` output. Local development may display the repository's placeholder version, but a published binary must display its actual release version and must not derive it from a private Harness package placeholder. Keep the existing startup timing and browser-opening behavior unchanged.

**Key interfaces:**
- The build-time Ready for Agent version currently used by the operator CLI's `--version` behavior should remain the single canonical version for the compiled distribution.
- Production Harness startup output should identify both the version and listening URL in its readiness log line.
- Shared application chrome should render the version with an unambiguous `v<semver>` presentation and remain usable on desktop and mobile.

**Acceptance criteria:**
- [ ] Starting a versioned production distribution emits a readiness line containing both `v<release-version>` and the Harness listening URL.
- [ ] The Harness UI visibly renders the same `v<release-version>` in shared chrome on every route.
- [ ] The value shown in startup output and the UI equals that distribution's `ready-for-agent --version` value.
- [ ] A published build never reports the private Harness package's placeholder version instead of the release version.
- [ ] Development builds have a deterministic version value and continue to start normally.
- [ ] Automated tests cover propagation of a non-placeholder version to the startup log and UI.
- [ ] Existing production startup ordering, browser opening, static asset serving, and CLI `--version` behavior remain intact.

**Out of scope:**
- Runtime update checks or notifications.
- Displaying Git commit hashes, build timestamps, or release channels.
- Changing release version calculation or semantic-versioning policy.

berenddeboer:
## Why

Operators need to know which Ready for Agent release is running without digging into package metadata or CLI help alone. Production startup and the shared UI chrome should surface the same build-time version as `ready-for-agent --version`.

## What

- Generate the launcher package version into both the operator binary and Harness modules before client build and embed.
- Include `v<semver>` in the production readiness log next to the listening URL.
- Render `v<semver>` in the shared root nav chrome on every route.
- Add tests for version generation, readiness log formatting, UI chrome wiring, and compiled-binary propagation.

Closes #369


berenddeboer:
### Status check handoff — `c621fa3fb108cf439c2576f849bfdd61f8072d61`

**Addressed**
- `main` / `ready-for-agent:knip`: unused export `READY_FOR_AGENT_VERSION_LABEL` in the launcher generated module.

**Changes**
- Emit `READY_FOR_AGENT_VERSION_LABEL` only in the Harness generated version module (used by UI chrome).
- Launcher module keeps `READY_FOR_AGENT_VERSION` only (CLI `--version`).
- Updated version-generation tests accordingly.

**Verification**
- `bunx nx run ready-for-agent:knip` passes
- `bunx nx run ready-for-agent:test` (version-related) passes
- Pre-commit affected lint/typecheck/test passed on commit

**Review feedback declined:** none

PR #440, opened 2026-07-24
fix(opencode): return nested /review task result, not parent resume

### LINKED ISSUE #439: Make OpenCode /review return the reviewer result without resuming build work
## Problem Statement

The Review Lifecycle Step invokes the required `/review` Agent Command and expects the Agent Turn Result to contain the reviewer's final assistant text. With OpenCode 1.18.3, `/review` runs as a child subtask and then synthetically resumes the parent build agent with `Summarize the task tool output above and continue with your task.` The OpenCode adapter returns the resumed parent's final text instead of the child reviewer's result.

This breaks the Review phase boundary. The child reviewer can correctly report findings without editing, while the resumed parent applies those findings before the lifecycle has classified them, run the controlled apply-findings pass, or rerun Pre-Commit. The parent may then return a marker valid for another phase, causing Review to fail.

Confirmed in Session `ses_06e8dc582ffeexfRF8oPcZzWjd` while working #416:

- Both `/review` child Sessions received the complete no-edit prompt.
- Both child reviewers made zero edits and ended with `READY_FOR_AGENT_RESULT: REVIEW_HAS_FINDINGS: medium`.
- After each child completed, OpenCode resumed the parent build agent.
- The parent applied each finding directly.
- On the second pass, the parent ended with `READY_FOR_AGENT_RESULT: REVIEW_FIXED`.
- The lifecycle correctly rejected that marker because it was still in the reviewing phase, producing `OpenCode did not report a unique final READY_FOR_AGENT_RESULT: REVIEW_CLEAN or REVIEW_HAS_FINDINGS: <low|medium|high>`.

## Solution

Make the OpenCode Agent Backend preserve the `/review` command boundary: the Agent Turn Result must expose the child reviewer's ordered final assistant text, and command execution must not let the resumed parent perform build work before the lifecycle interprets the review outcome.

Keep the existing strict phase-specific result protocols. Do not reinterpret `REVIEW_FIXED` as a reviewing result or merely relax the verdict fallback, because by that point uncontrolled edits may already have bypassed the Review Fix Round and nested Pre-Commit sequence.

If OpenCode's current `--command` behavior cannot directly provide this observable contract, use another OpenCode invocation strategy that does. The Work Item must retain its durable parent Session identity and later build-model turns must still continue that Session.

## User Stories

1. As an operator, I want a correct `/review` result to advance through the Review state machine, so that valid reviews do not fail because output from another agent was captured.
2. As an operator, I want Review Findings applied only in a controlled Review Fix Round, so that remediation is followed by the required Pre-Commit and rerun policy.
3. As a reviewer, I want my no-edit instruction to remain effective for the whole reviewing pass, so that reviewing cannot mutate the worktree indirectly through a resumed parent.
4. As a lifecycle maintainer, I want Agent Command results to identify the command's actual final response, so that phase-specific machine protocols remain trustworthy.
5. As a backend maintainer, I want deterministic coverage of nested command output, so that future OpenCode changes cannot silently switch the result back to resumed-parent text.

## Implementation Decisions

- Define and enforce the Agent Command result boundary in the OpenCode adapter rather than weakening Review's marker parser.
- Preserve the Work Item's existing durable Session ID across review and later build-model turns.
- Return only the command reviewer's ordered final assistant text as the reviewing Agent Turn Result.
- Prevent or isolate any automatic parent continuation that could edit the shared worktree during the reviewing pass.
- Keep `REVIEW_CLEAN` and `REVIEW_HAS_FINDINGS: <low|medium|high>` as the only valid reviewing outcomes.
- Keep apply-findings outcomes, including `REVIEW_FIXED`, valid only during the build-model apply pass.
- Preserve the current severity, Review Fix Round, nested Pre-Commit, and risk-based rerun behavior.
- Add no compatibility path that treats wrong-phase markers as successful reviews.
- No persistence or GraphQL schema change is required.

## Testing Decisions

- Add a deterministic regression at the OpenCode Agent Backend seam, which is the highest seam that owns command invocation and output normalization.
- Model the observed case where a `/review` child ends with `REVIEW_HAS_FINDINGS: medium` while a resumed parent later emits `REVIEW_FIXED`; assert that the normalized command result is the child finding and excludes the parent continuation.
- Assert that the reviewing command cannot produce parent worktree edits before the lifecycle starts its apply-findings pass.
- Exercise the lifecycle with the normalized review result and verify that findings enter a Review Fix Round, followed by the existing nested Pre-Commit and rerun policy.
- Retain strict malformed, duplicate, non-final, and wrong-phase marker tests.
- Extend the opt-in OpenCode integration coverage to validate `/review` command result identity when practical; normal CI must remain deterministic and must not require a live model.

## Out of Scope

- Accepting `REVIEW_FIXED` during a reviewing pass.
- Redesigning review severity or the Review Fix Round policy.
- Changing the generic `READY_FOR_AGENT_RESULT` protocols.
- Depending on OpenCode's private database as a production result transport.
- Adding support for legacy command behavior after the corrected contract is implemented.

## Further Notes

This is not reviewer noncompliance. Transcript inspection confirmed that the child reviewers followed their prompts exactly. The failure is caused by the harness treating OpenCode's resumed parent response as the `/review` result.

berenddeboer:
## Summary

- Preserve the OpenCode `/review` Agent Command boundary by returning the nested child `task_result` as the Agent Turn Result instead of the automatic parent resume text.
- Stop parent continuation after the command task completes so reviewing cannot edit the worktree before the controlled Review Fix Round.
- Fail closed when the parent process cannot be terminated after a command result is captured.
- Add deterministic OpenCode adapter and Review lifecycle regressions for the observed nested-command failure mode.

Closes #439.

## Test plan

- [x] `bunx nx run opencode:test`
- [x] `bunx nx run opencode:typecheck`
- [x] `bunx nx run opencode:lint`
- [x] `bunx nx run work-item-lifecycle:test`
- [x] `bunx nx run work-item-lifecycle:typecheck`
- [x] `bunx nx run work-item-lifecycle:lint`
- [x] Pre-commit `nx affected` checks passed on commit


berenddeboer:
## Status check follow-up

**Commit:** `719c68901e4a6a10d1bf273fdbfc75b25d2445b1`

### Cause
`main` failed on `opencode:test`: the regression that expected a command turn to reject when the fake CLI exited immediately after the nested task was racy. On GHA, `kill` often still succeeded against a process that had not fully reaped, so the promise resolved instead of rejected.

### Changes
- Kill the OpenCode process after a nested command result only when `handle.isRunning` is true.
- If the process has already exited, accept the command `task_result` without racing `kill`.
- Replace the flaky “must reject on immediate exit” test with a deterministic immediate-exit success case that still asserts command-result identity.

### Verification
- `bunx nx run opencode:test`
- `bunx nx run opencode:lint`
- `bunx nx run opencode:typecheck`
- Pre-commit `nx affected` (lint/typecheck/test) passed on commit

### Review feedback
- **Addressed:** prior review ask to fail closed when parent isolation cannot be established via kill — kill still fails the turn when the process is running and kill fails; already-exited processes no longer trip a flaky kill race.
- **Declined:** none

PR #847, opened 2026-08-06
fix(work-item-lifecycle): accept trailing READY_FOR_AGENT_RESULT marker amid prose

### LINKED ISSUE #845: Accept a valid trailing READY_FOR_AGENT_RESULT marker even amid explanatory prose
## Problem

`uniqueFinalResultLine` in `packages/work-item-lifecycle/src/lib/review.ts` (used by `parseReviewResult`, `parseApplyReviewResult`, and `parseRerunAssessmentResult`) rejects the result unless the `READY_FOR_AGENT_RESULT: ...` marker is **both**:

1. the only line anywhere in the output matching `/^READY_FOR_AGENT_RESULT:/i`, and
2. the very last non-empty line of the output.

```ts
const uniqueFinalResultLine = (output: string): string | null => {
  const lines = output.split("\n").map((line) => line.trim())
  const nonEmptyLines = lines.filter((line) => line !== "")
  const resultLines = lines.filter((line) =>
    /^READY_FOR_AGENT_RESULT:/i.test(line),
  )
  const finalLine = nonEmptyLines.at(-1)

  if (
    resultLines.length !== 1 ||
    finalLine === undefined ||
    finalLine !== resultLines[0]
  ) {
    return null
  }
  return finalLine
}
```

When a reviewing pass returns a valid, unambiguous, well-formed marker but it isn't in that exact position (e.g. explanatory prose trails it, or an earlier draft/attempt also happened to contain a marker-shaped line), `parseReviewResult`/`parseApplyReviewResult` return `null` even though a correct answer was clearly reported. The caller (`review.ts` around line 505) then either burns an extra Agent Turn re-asking for a bare verdict (`buildReviewVerdictPrompt`, ~line 507) or fails the Review Step Run outright with "did not report a unique final READY_FOR_AGENT_RESULT: ..." (~line 530, ~line 574).

### Observed case

Session `ebb1b7a2-0852-4989-a394-8beaee349cda` (Work Item 838): the initial `/review` turn returned early with "Standards review is in. Waiting on the Spec review before aggregating." (no marker yet, since it was waiting on an async sub-agent). The harness correctly detected no result line and issued the classify follow-up prompt. On resume, the pending sub-agent result arrived, and the model aggregated both reviews into a full report (Standards findings, Spec findings, a summary paragraph) and correctly ended with:

```
READY_FOR_AGENT_RESULT: REVIEW_HAS_FINDINGS: medium
```

The line was well-formed and the only occurrence, but the Step Run still failed with "did not report a unique final READY_FOR_AGENT_RESULT: REVIEW_CLEAN or REVIEW_HAS_FINDINGS: <low|medium|high>" — the report text around the marker was enough to trip up the strict final-line/uniqueness check in at least some execution paths through this logic.

## Proposal

Make the three parsers tolerant of explanatory prose and multiple candidate lines, as long as a valid, well-formed marker can be found unambiguously:

- Scan all lines matching `/^READY_FOR_AGENT_RESULT:/i`, not just the last non-empty line.
- Try to parse each candidate against the known valid patterns for that pass type (`REVIEW_CLEAN`, `REVIEW_HAS_FINDINGS: <low|medium|high>`, etc.) — ignore lines that don't parse (they aren't valid markers, e.g. a stray mention in prose).
- If one or more lines parse successfully, take the **last** one that parses (models correct/refine forward, so the latest valid statement wins over an earlier draft).
- Only return `null` when *no* line parses to a valid, known marker — i.e. keep failing the Step Run for a genuinely missing or malformed result, just stop failing it for a *found and valid* one that merely wasn't isolated on its own final line.

This removes the need for the extra `continueTurn` round-trip purely to re-elicit a bare marker when a valid one was already present in the first response, saving tokens/turns/time on every reviewing, apply-findings, and rerun-assessment pass that ends this way.

## Where to change

- `packages/work-item-lifecycle/src/lib/review.ts`: `uniqueFinalResultLine` and its three callers (`parseReviewResult`, `parseApplyReviewResult`, `parseRerunAssessmentResult`).
- `packages/work-item-lifecycle/test/review.spec.ts`: update the existing "rejects missing, duplicate, non-final, ..." test cases (around lines 195, 260, 355) to match the new intended behavior — keep coverage for "no valid marker anywhere" still returning `null`, but change the duplicate/non-final cases to expect the last valid marker to be accepted.

## Non-goals

Don't loosen this so far that it accepts garbage: a candidate line must still match one of the known, fully-formed result patterns (correct keyword, valid severity value, non-empty reason where required). Only the "must be the sole occurrence" and "must be the literal final line" constraints should be relaxed.

berenddeboer:
## Why

The reviewing, apply-findings, and rerun-assessment parsers in `review.ts` required the
`READY_FOR_AGENT_RESULT: ...` marker to be *both* the sole matching line in the output *and* the
literal final non-empty line. When a reviewing pass produced a well-formed, unambiguous marker but
trailed it with explanatory prose (or an earlier draft-shaped line also happened to match the
marker prefix), the Step Run either burned an extra Agent Turn re-asking for a bare verdict or
failed outright with "did not report a unique final READY_FOR_AGENT_RESULT: ..." — even though a
correct answer was clearly present. This was observed in Work Item 838, where a valid
`REVIEW_HAS_FINDINGS: medium` marker was rejected because the aggregated report included prose
around it.

## What changed

- Replaced the strict `uniqueFinalResultLine` helper with `candidateResultLines` + a generic
  `lastValidResult(output, tryParseLine)` helper that scans every line matching
  `/^READY_FOR_AGENT_RESULT:/i`, tries to parse each against the known valid patterns, and keeps
  the **last** one that parses successfully. Lines that don't parse (stray mentions, malformed
  severities, foreign markers) are ignored rather than causing an outright rejection.
- Extracted per-line matching logic into `tryParseReviewLine`, `tryParseApplyReviewLine`, and
  `tryParseRerunAssessmentLine`, each driven through `lastValidResult` from `parseReviewResult`,
  `parseApplyReviewResult`, and `parseRerunAssessmentResult` respectively.
- `null` is now returned only when *no* candidate line parses to a valid, known marker — the "sole
  occurrence" and "must be the final line" constraints are gone, but a candidate must still fully
  match one of the known result patterns (correct keyword, valid severity, non-empty reason where
  required), so garbage or foreign markers are still rejected.
- Updated the two `ReviewResultError` messages from "did not report a unique final ..." to "did not
  report a valid ..." to match the relaxed semantics, and corrected a now-stale comment in the
  rerun-assessment branch that referenced duplicate/ambiguous markers forcing a conservative
  rerun.

## Verification

- Updated `review.spec.ts` unit tests for all three parsers: cases with duplicate or non-final valid
  markers now assert the last valid one is accepted; cases with no valid marker anywhere still
  assert `null`.
- Updated the two `review()` integration tests that previously asserted failure on duplicated/non-
  final markers to instead assert the Step Run now succeeds, accepting the last valid marker (one
  exercises the duplicate-markers path landing on `REVIEW_HAS_FINDINGS` → apply →
  `REVIEW_CLEARED`; the other exercises a marker followed by trailing prose landing on
  `REVIEW_CLEAN`).
- `bun test` for `work-item-lifecycle`: 560 pass / 0 fail.
- `nx typecheck` and `nx lint` for `work-item-lifecycle`: both clean.

Closes #845

### DIFF
--- a/apps/harness/e2e/steps/catalog-only-agent-model.ts
+++ b/apps/harness/e2e/steps/catalog-only-agent-model.ts
@@ -146,7 +146,8 @@ Given(
 
 Given(
   "the Harness runs Claude Code with the {string} Agent Model",
-  async (_fixtures, model: string) => {
+  // biome-ignore lint/correctness/noEmptyPattern: playwright-bdd requires the first argument to use the object destructuring pattern, even when no fixtures are needed.
+  async ({}, model: string) => {
     await seedClaudeHarnessDefault(model)
   },
 )
--- a/packages/work-item-lifecycle/src/lib/review.ts
+++ b/packages/work-item-lifecycle/src/lib/review.ts
@@ -174,28 +174,35 @@ export const buildRerunAssessmentPrompt = () =>
     "when a full reviewing pass is required.",
   ].join("\n")
 
-const uniqueFinalResultLine = (output: string): string | null => {
-  const lines = output.split("\n").map((line) => line.trim())
-  const nonEmptyLines = lines.filter((line) => line !== "")
-  const resultLines = lines.filter((line) =>
-    /^READY_FOR_AGENT_RESULT:/i.test(line),
-  )
-  const finalLine = nonEmptyLines.at(-1)
+const candidateResultLines = (output: string): string[] =>
+  output
+    .split("\n")
+    .map((line) => line.trim())
+    .filter((line) => /^READY_FOR_AGENT_RESULT:/i.test(line))
 
-  if (
-    resultLines.length !== 1 ||
-    finalLine === undefined ||
-    finalLine !== resultLines[0]
-  ) {
-    return null
+/**
+ * Try each candidate READY_FOR_AGENT_RESULT line (in order) against
+ * `tryParseLine` and keep the last one that parses to a valid, known
+ * marker. Explanatory prose around or between candidates, and earlier
+ * candidates that don't parse (or are superseded by a later valid one),
+ * are ignored. Returns null only when no candidate parses.
+ */
+const lastValidResult = <T>(
+  output: string,
+  tryParseLine: (line: string) => T | null,
+): T | null => {
+  let result: T | null = null
+  for (const line of candidateResultLines(output)) {
+    const parsed = tryParseLine(line)
+    if (parsed !== null) {
+      result = parsed
+    }
   }
-  return finalLine
+  return result
 }
 
 const hasResultLine = (output: string): boolean =>
-  output
-    .split("\n")
-    .some((line) => /^READY_FOR_AGENT_RESULT:/i.test(line.trim()))
+  candidateResultLines(output).length > 0
 
 const parseSeverity = (raw: string): ReviewSeverity | null => {
   const value = raw.trim().toLowerCase()
@@ -215,23 +222,12 @@ const parseDeferredSeverity = (raw: string): DeferredReviewSeverity | null => {
 
 const boundReason = (reason: string): string => reason.trim().slice(0, 500)
 
-/**
- * Parse the unique final READY_FOR_AGENT_RESULT line from a reviewing pass.
- * Returns null for missing, duplicate, non-final, or unrecognized markers.
- */
-export const parseReviewResult = (
-  output: string,
-): ReviewingPassResult | null => {
-  const finalLine = uniqueFinalResultLine(output)
-  if (finalLine === null) {
-    return null
-  }
-
-  if (/^READY_FOR_AGENT_RESULT:\s*REVIEW_CLEAN$/i.test(finalLine)) {
+const tryParseReviewLine = (line: string): ReviewingPassResult | null => {
+  if (/^READY_FOR_AGENT_RESULT:\s*REVIEW_CLEAN$/i.test(line)) {
     return { _tag: "clean" }
   }
 
-  const hasFindings = finalLine.match(
+  const hasFindings = line.match(
     /^READY_FOR_AGENT_RESULT:\s*REVIEW_HAS_FINDINGS\s*:\s*(.+)$/i,
   )
   if (hasFindings?.[1] !== undefined) {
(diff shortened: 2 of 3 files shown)


berenddeboer:
Fixed the failing `harness` check (Actions job [92721044172](https://github.com/berenddeboer/ready-for-agent/actions/runs/31131463152/job/92721044172)) in commit `a47dfd50c4584f27504def56f49dc6152bab0af2`.

**Root cause**: `bunx bddgen` (playwright-bdd's test generator) throws `Error: First argument must use the object destructuring pattern` for the `Given` step at `apps/harness/e2e/steps/catalog-only-agent-model.ts:149` ("the Harness runs Claude Code with the {string} Agent Model"). playwright-bdd inspects each step callback's source text and requires the first parameter to be an object destructuring pattern (even an empty `{}`) to extract requested fixtures; the step used a plain identifier (`_fixtures`) instead, which aborted test generation before any spec files were produced — failing the whole `harness:e2e` target (both this PR's `harness` job and the same live e2e job on `main`, run from commit 9d17672 which introduced the step).

**Fix**: changed the first parameter to `{}` (matching the sibling `Given` step immediately above it that also needs no fixtures), and added a `biome-ignore lint/correctness/noEmptyPattern` comment since biome otherwise flags the now-required empty pattern.

**Verification**:
- `bunx bddgen` in `apps/harness` now exits 0 (previously threw and aborted with exit 1).
- `bunx playwright test --list` lists all 11 tests across 3 feature files, including the previously-unreachable `catalog-only-agent-model.feature.spec.js` scenarios.
- `nx run harness:lint` and `nx run harness:typecheck`: clean.
- `nx run harness:test`: 475 + 92 pass, 0 fail (sanity check; unrelated to e2e generation).
- Did not run the live Playwright e2e suite locally (needs the vault-backed `E2E_KEYMAXXER_MASTER_KEY` secret, only available in CI); verified at the exact point the CI job failed (test generation). This push will trigger a fresh CI run to confirm end-to-end.

**Review feedback addressed**: none — no automated review run or comment existed on this PR at the time of this check.

**Review feedback declined**: none.

PR #848, opened 2026-08-06
fix(harness): prevent Harness test runs from prompting for Keymaxxer

### LINKED ISSUE #844: Prevent Harness test runs from prompting for Keymaxxer
## Problem Statement

Harness test commands can inherit an environment in which Keymaxxer is available on `PATH`. Because Keymaxxer is automatically enabled unless `KEYMAXXER_ENABLED` is exactly `false`, unattended agent runs can start Keymaxxer and display unlock or approval prompts to the operator.

This is especially disruptive when agents run tests repeatedly. Ordinary unit, slow, and smoke tests do not need operator credentials and must never prompt. The live Harness end-to-end suite is different: it intentionally validates the production application, Keymaxxer Sidecar, command-line client, and controlled Forge Repositories together. Setting `KEYMAXXER_ENABLED=false` for that suite would invalidate its contract, and its direct Keymaxxer CLI calls do not honor that variable.

## Solution

Make every supported Harness test entry point non-interactive by default:

- Force `KEYMAXXER_ENABLED=false` for the Harness unit, slow-test, and smoke Nx targets, overriding any inherited value.
- Apply the same disabled value in the unit-test runner so direct invocation of the supported runner is also safe.
- Keep live e2e Keymaxxer-enabled because credential brokering is part of the scenario.
- Require a fixture vault master key for non-interactive live e2e runs. Configure the fixture environment to unlock without a dialog and deny approval prompts.
- When no fixture master key is available, fail before starting the Harness, Sidecar, or any direct Keymaxxer CLI command.
- Preserve ambient-vault local e2e only behind the explicit human opt-in `E2E_ALLOW_KEYMAXXER_PROMPTS=1`.
- Keep existing CI behavior: trusted runs with `E2E_KEYMAXXER_MASTER_KEY` execute live e2e, while workflow-level fork handling may continue to skip the target when repository secrets are unavailable.

## User Stories

1. As a developer delegating test execution to an agent, I want ordinary Harness tests to disable Keymaxxer automatically, so that repeated test runs never interrupt me with unlock dialogs.
2. As an agent running the Harness unit suite through Nx, I want the target configuration to override an inherited enabled value, so that the result is deterministic across environments.
3. As a developer invoking the supported unit-test runner directly, I want the runner to disable Keymaxxer itself, so that bypassing Nx does not reintroduce prompts.
4. As a developer running the slow production SSE test, I want Keymaxxer disabled at the target boundary, so that production lifecycle availability detection cannot launch a keyholder unnecessarily.
5. As a developer running the development smoke test, I want its no-Keymaxxer contract declared at the target boundary as well as enforced by its child environment, so that the command remains visibly and reliably non-interactive.
6. As a CI operator, I want live e2e to continue exercising the real Keymaxxer Sidecar and fixture vault, so that credential-backed production behavior remains covered.
7. As an agent lacking the fixture master key, I want live e2e to fail immediately with an actionable diagnostic, so that it cannot fall back to the operator's ambient vault or display a prompt.
8. As a local human intentionally testing against my ambient vault, I want an explicit prompt opt-in, so that interactive behavior remains possible but can never happen accidentally.
9. As a contributor, I want the non-interactive policy centralized across e2e server startup and fixture cloning, so that direct Keymaxxer CLI calls cannot bypass the guard.
10. As a maintainer, I want tests that lock down the ordinary-target environment and e2e preflight behavior, so that future target or process changes do not restore prompts.
11. As a maintainer, I want Nx caching to remain deterministic, so that enabled and disabled executions cannot silently share an ambient-environment-dependent cache entry.
12. As a developer reading the Harness documentation, I want the safe default, fixture-key requirement, and explicit interactive opt-in documented, so that I know which command can access my vault before running it.

## Implementation Decisions

- The Nx target environment is the authoritative boundary for supported Harness test commands. Unit, slow-test, and smoke targets set `KEYMAXXER_ENABLED` to the string `false` rather than relying on callers or CI workflows.
- The unit-test shell runner also exports `KEYMAXXER_ENABLED=false` to cover direct supported invocation and all Bun/Vitest children.
- Development and production start targets remain unchanged; operator-controlled Keymaxxer behavior outside tests is not affected.
- Live e2e remains a Keymaxxer integration and explicitly runs with Keymaxxer enabled. It is not converted into a credential-free or mocked suite.
- A shared e2e environment-policy function distinguishes fixture-vault mode from explicitly opted-in interactive mode. Missing credentials without opt-in are an error.
- `E2E_KEYMAXXER_MASTER_KEY` remains the preferred fixture key, with the existing `KEYMAXXER_MASTER_KEY` fallback retained.
- CI or explicitly requested fixture-vault mode without a master key always fails closed. Interactive opt-in must not weaken CI's non-interactive guarantee.
- Fixture mode continues to use an isolated temporary home, the checked-in encrypted vault, `KEYMAXXER_MASTER_KEY`, and `KEYMAXXER_APPROVE=deny`.
- The fail-fast check occurs before starting the production Harness or reaching fixture cloning, preventing both Sidecar-mediated and direct CLI prompts.
- Ambient local-vault behavior is available only when `E2E_ALLOW_KEYMAXXER_PROMPTS=1` is explicitly set.
- Existing fork-PR workflow skipping remains a workflow concern and is not replaced by a successful no-op inside the e2e target.
- Documentation and the live-e2e ADR are updated because the previous documented local default allowed normal Keymaxxer prompts.

## Testing Decisions

- Test external policy at the highest practical seams rather than asserting incidental implementation text.
- Extend Harness topology/configuration coverage to verify that the resolved ordinary test targets force `KEYMAXXER_ENABLED=false` and that live e2e remains enabled.
- Unit-test the shared e2e policy for fixture-key selection, the legacy master-key fallback, explicit interactive opt-in, default rejection, and fail-closed CI/fixture modes.
- Add a process-level regression check showing that live e2e startup without a fixture key or interactive opt-in exits promptly with the expected diagnostic before any server or Keymaxxer process can start.
- Verify that an inherited `KEYMAXXER_ENABLED=true` cannot override the ordinary test targets.
- Run the Harness unit and slow-test targets after implementation.
- Run the Harness smoke target to verify its application child still starts with Keymaxxer disabled.
- Verify live e2e startup fails fast without credentials. Run the complete live e2e only when the fixture master key is available; never use the operator's ambient vault during automated verification.

## Out of Scope

- Removing Keymaxxer from the live end-to-end architecture.
- Mocking Forge credentials or changing the controlled GitHub/GitLab fixture scenarios.
- Changing Keymaxxer's own unlock, approval, or environment-variable behavior.
- Disabling Keymaxxer for normal Harness development or production startup.
- Changing repository-secret provisioning, fixture-vault regeneration, or fork pull-request workflow policy.
- Guaranteeing behavior for arbitrary direct invocations that bypass every supported Nx target and test runner.

## Further Notes

Only a trimmed, case-insensitive exact value of `false` currently disables Keymaxxer; values such as `0` or `no` do not. Use the literal string `false` in target and runner environments.

A blanket workspace or CI-job variable is insufficient: it would obscure target intent, interact poorly with cached tests, and break the live e2e contract. The distinction between credential-free test targets and non-interactive credential-backed e2e must remain explicit.

berenddeboer:
## Why

Keymaxxer auto-enables on any Harness test command unless `KEYMAXXER_ENABLED`
is exactly `"false"`. When Keymaxxer happens to be on `PATH` (as it commonly
is for agents/operators), that meant unit tests, the slow SSE test, and the
dev smoke test could silently trigger unlock or approval prompts on
repeated/unattended runs. Live e2e is the one suite that legitimately needs
Keymaxxer enabled (it exercises the real Sidecar, CLI, and fixture
credentials), but it had no fail-fast guard: without a fixture master key it
silently fell back to the operator's ambient vault and could prompt too.

## What changed

- **Ordinary test targets are now non-interactive by default.**
  `apps/harness/project.json`'s `test`, `slow-test`, and `smoke` Nx targets
  set `env.KEYMAXXER_ENABLED: "false"`, which overrides any inherited value
  (`nx:run-commands`' `env` option wins over ambient `process.env`).
- **The unit-test runner enforces the same default.**
  `scripts/run-unit-tests.sh` now exports `KEYMAXXER_ENABLED=false`
  unconditionally, so direct/CI invocation of the runner stays safe even
  outside the Nx target.
- **Live e2e keeps Keymaxxer enabled but fails closed.** A new shared policy
  module, `apps/harness/e2e/support/keymaxxer-e2e-policy.ts`, is the single
  seam both `start-live-harness.ts` (the Playwright `webServer`) and
  `clone-fixture-repo.ts` (the CLI clone helper) call before touching
  Keymaxxer:
  - A fixture master key (`E2E_KEYMAXXER_MASTER_KEY`, falling back to legacy
    `KEYMAXXER_MASTER_KEY`) always selects the checked-in fixture vault
    (`KEYMAXXER_APPROVE=deny`, isolated temp `HOME`).
  - CI or an explicit fixture-vault request (`E2E_USE_FIXTURE_VAULT=1`)
    without a key fails closed with a diagnostic.
  - Otherwise, `E2E_ALLOW_KEYMAXXER_PROMPTS=1` is the only way to opt into
    the operator's ambient `~/.keymaxxer` vault.
  - Anything else throws before any temp dir, fake CLI, production-build
    check, or Keymaxxer process is touched.
- **Docs updated** (`apps/harness/README.md`,
  `docs/adr/0021-live-harness-end-to-end-test.md`) to describe the new
  default and the interactive opt-in, replacing the previous "local runs
  prompt normally" language.

## Verification

- `nx run harness:test` — 490 bun tests + 92 vitest tests pass.
- `nx run harness:slow-test` and `nx run harness:smoke` pass.
- `nx run harness:lint`, `nx run harness:typecheck`, and
  `nx affected -t lint knip typecheck test` pass.
- New coverage: `test/keymaxxer-e2e-policy.test.ts` (policy matrix:
  master-key selection, legacy fallback, interactive opt-in, default
  rejection, CI/fixture-vault fail-closed), `test/start-live-harness-preflight.test.ts`
  (process-level regression proving live e2e exits fast with the diagnostic,
  before ever checking for a production build), and extended
  `test/topology.test.ts` assertions (target-level env overrides, `e2e`
  target left untouched, and a live regression proving the runner's own
  `export KEYMAXXER_ENABLED=false` line wins over an inherited
  `KEYMAXXER_ENABLED=true`).
- Manually exercised all three live-e2e policy paths by invoking
  `start-live-harness.ts` directly: no-credential fail-fast, CI-without-key
  fail-closed even with the interactive opt-in set, and interactive opt-in
  proceeding past the policy check (only failing later on the unrelated
  missing production build).
- Out of scope (per issue): removing Keymaxxer from live e2e, mocking Forge
  credentials, changing Keymaxxer's own unlock/approval behavior, or
  altering fork-PR workflow skipping.

Closes #844


berenddeboer:
Pushed 3a14dc4 to fix the failing "harness" check (Actions job [92721628394](https://github.com/berenddeboer/ready-for-agent/actions/runs/31131652541/job/92721628394)).

**Root cause:** the `harness:e2e` target failed during `bddgen` (test generation), before Playwright ever ran, so both the smoke-adjacent and live-e2e checks in the `harness` job never got a chance to execute. The catalog-only-Agent-Model step `"the Harness runs Claude Code with the {string} Agent Model"` (added in #846, already on `main`, unrelated to this PR's Keymaxxer changes) declared its first parameter as a plain identifier `_fixtures` instead of an object-destructuring pattern. `playwright-bdd`'s `bddgen` inspects step-function source text and throws `First argument must use the object destructuring pattern` when the first parameter isn't literally `{ ... }`.

**Fix:** `apps/harness/e2e/steps/catalog-only-agent-model.ts` — changed the step to `async ({}, model: string) => { ... }`, matching every other fixture-less step in the file, with a `biome-ignore lint/correctness/noEmptyPattern` (biome doesn't know playwright-bdd requires that literal shape).

**Verification:**
- `bddgen` now exits 0 (previously crashed with the destructuring error).
- `playwright test --list` resolves all 11 scenarios across the 3 feature files (previously crashed before listing anything).
- `nx run harness:lint`, `nx run harness:typecheck`, and `nx run harness:test` (490 bun + 92 vitest tests) all pass.

**Review feedback addressed:** none — no automated-review run or comment existed on this PR at the time of this handoff (only the `PR` workflow's `harness` job, which is CI, not a review).

Issue #933, opened 2026-08-10
First-run feedback (v0.18.0, Linux arm64): `add` before `start` reads as a network failure, and `paused: true` is undocumented


cgbarlow:
Fresh install on a devcontainer (Node v22.23.2, npm 10.9.8, Linux arm64). Install, PATH, `add`, and forge inference all worked. Three bits of first-run friction, in order of how much they cost me.

## 1. `add` before `start` surfaces as a connectivity error

The quick start in the README lists:

```
npx ready-for-agent@latest
...
ready-for-agent add /path/to/local/repo
```

Nothing there says the harness has to be running for `add` to work, so I ran `add` first:

```
$ ready-for-agent add /workspaces/workspace-basic/braintrust
[01:46:37.386] ERROR (#5): GraphqlRequestFailed: Unable to connect. Is the computer able to access the url?

Start the Harness first: ready-for-agent start (or bun run ready-for-agent start)

Start the Harness first: ready-for-agent start (or bun run ready-for-agent start)
    at catch (/$bunfs/root/ready-for-agent:128895:53)
    at ~effect/Utils/internal (/$bunfs/root/ready-for-agent:1259:20)
    at failWithCatch (/$bunfs/root/ready-for-agent:6409:34)
    at processTicksAndRejections (native:7:39)
    at GraphqlApi.addRepository (/$bunfs/root/ready-for-agent:208003:50)
    at GraphqlApi.addRepository (definition) (/$bunfs/root/ready-for-agent:128866:46)
    at Cli.addRepository (/$bunfs/root/ready-for-agent:127152:35)
    at Cli.addRepository (definition) (/$bunfs/root/ready-for-agent:207994:44)
```

The correct remedy *is* in there, which I appreciate. But it's third in reading order, behind a headline that points at the network — `Unable to connect. Is the computer able to access the url?` sends you to check firewalls and proxies, not to start a local service. Then it's printed twice, and followed by an 8-frame stack trace of bundled internals for what is an expected, fully-handled condition.

Suggestion, roughly in order of value:

- Lead with the actual cause: `Harness not running at http://127.0.0.1:6056 — start it with: ready-for-agent start`. Drop the GraphQL/connectivity framing when the target is loopback and nothing is listening.
- Print the remedy once.
- Suppress the stack trace for this case (or put it behind `--log-level debug`).
- Optional, and the nicest outcome: have `add` start the harness itself, or offer to.

## 2. README quick start doesn't state the ordering

[...]

PR #934, opened 2026-08-10
Add a CLI loop to enrol every relevant issue

cgbarlow:
@berenddeboer Follow-up to #933, where I mentioned running the harness headless.

## The gap

README says:

> Currently the harness does not automatically pick issues to work on. Click on the kebab menu and implement an issue end to end via "Implement now".

That's the right default when someone is watching. But it leaves no path for a prepared batch of tickets on a headless box or from cron — the click is the only way in, and `ready-for-agent` the binary only exposes `start` and `add`.

## What this adds

`scripts/drain-ready-issues.sh` — enrols every open Relevant Issue that has no unfinished Work Item, then follows the Work Items until they settle. It drives the same GraphQL API the UI calls, so it takes no shortcuts around the lifecycle: `updateRepositorySettings`, `implementNow`, `queue`, `implementAllWithAutoMerge`.

```bash
scripts/drain-ready-issues.sh --repo owner/name --dry-run  # show the plan
scripts/drain-ready-issues.sh --repo owner/name            # enrol, then follow
scripts/drain-ready-issues.sh --repo owner/name --status   # follow an existing run
scripts/drain-ready-issues.sh --repo owner/name --no-retry  # report failures, retry nothing
```

Design decisions worth your eye, since they encode assumptions about the domain:

- **Blockers stay yours.** An issue with open blockers is admitted with `queue`, not `implementNow`, so the harness keeps ordering rather than the script second-guessing it.
- **Parent issues are skipped** unless `--auto-merge` is passed, on the reading that a parent is never implemented directly and `implementAllWithAutoMerge` is the only route to Merge Mode ALWAYS on children. Children are admitted on their own merit otherwise.
- **Settings are read back and rewritten unchanged** when the caller didn't ask to change them, so it can't silently clear a per-repo backend or model override.
- **Failures are retried, and explained.** A failed build isn't terminal — the harness parks it with `canRetry` and waits for a human to press Retry, which is exactly where an unattended drain stops for good. So the script presses it, up to `--max-retries` (default 3), and reads the reason out of the coder's own transcript because `statusMessage` gives only "Claude Code failed to implement the Work Item issue". A spent agent quota is treated as one wall for all Work Items rather than N separate failures.
- **Needs Human is surfaced, not hidden.** Auto-merge only clears low-risk PRs, so an unattended run is *expected* to park some items; the poll output counts them rather than implying everything sailed through.
- Honours `READY_FOR_AGENT_GRAPHQL_URL`, matching `ready-for-agent add` on non-default ports.

## Testing

Run against a live harness (v0.18.0, Linux arm64, GitHub repo, 10 relevant issues with a real blocker DAG):

- `--dry-run` routed 6 unblocked issues to `implementNow` and 4 blocked ones to `queue`, correctly
- `--status` against a repo with no Work Items exits cleanly rather than polling forever
- unreachable harness, unknown `--repo`, multiple repos with no `--repo`, and unknown flags all exit non-zero with an actionable message
- `bash -n` clean; no shellcheck config in the repo so I matched the existing style in `scripts/*.sh` instead

The interactive setup (backend, build model, Thinking Level) was exercised from a reset harness: prompts appear on first run, the choices are written to the Repository, and a second run reads them back without asking again. `--backend`, `--build-model` and `--thinking-level` are validated against the live catalog before anything is written, and an unattended run with no model set stops with the options rather than picking one.

The retry path was exercised the hard way, by an actual batch failure rather than a simulated one. All 6 Work Items admitted by the script failed at Build inside thirteen seconds; every transcript ended on `You've hit your session limit`. Retrying moved all 6 off Failed, two reached Review, and one hit the wall again — the hold parsed the new reset time, held 4h23m, and resumed on its own. A later run caught the hold reading that clock against the current time rather than against the failure it came from, so a drain that outlived its own reset idled 5h20m on a quota it already had back; the fix anchors the clock to the failure's timestamp, and the timestamp now returns on stdout because a global set inside `$( )` never reached the caller. `stalled`, the quota parser and the transcript reader also have direct unit checks against real transcripts and synthetic Work Item sets.

**A Work Item has now been observed end to end.** Work Item #202 reached Complete with its PR merged under `--auto-merge`, so the terminal states past Review are observed rather than inferred from the schema. The batch behind it drained on the same run: 6 Work Items retried, PRs opened, status checks run. What remains unobserved is the unhappy end of that path — a PR that fails its checks, and `waitForReadyForReviewChecks` holding it. Worth a maintainer's eye before anyone trusts it fully unattended.

Happy to change the name, the flags, or drop the README section if you'd rather this stay undocumented.

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,6 +158,24 @@ Pick the "Implement locally" option to implement the issue in the new
 worktree, but without creating a PR yet. This allows you to test and
 verify before a commit or PR exists.
 
+### Enrolling every issue from the command line
+
+If you have prepared a batch of tickets and want them all admitted
+without clicking through each one — on a headless box, or from cron —
+`scripts/drain-ready-issues.sh` does that over the same GraphQL API the
+UI uses:
+
+```bash
+scripts/drain-ready-issues.sh --repo owner/name --dry-run  # show the plan
+scripts/drain-ready-issues.sh --repo owner/name            # enroll, then follow
+```
+
+Issues with open blockers are admitted with Queue rather than Implement
+Now, so the harness still decides the order. Add `--auto-merge` to turn
+on auto-merge for the repo; the AI risk assessment still applies, so
+higher-risk PRs continue to stop at Needs Human. Requires `curl` and
+`jq`.
+
 ## Configuration
 
 ### Stop opening a browser window
--- a/scripts/drain-ready-issues.sh
+++ b/scripts/drain-ready-issues.sh
@@ -20,6 +20,15 @@
 # is what an unattended run must do — rather than choose a model, and a price,
 # on your behalf, the script stops and tells you the options.
 #
+# A failed build is not terminal. The harness parks the Work Item with canRetry
+# set and waits for someone to press Retry, which on an unattended run is where
+# a drain stops for good — so failures are retried here, up to --max-retries.
+# The reason is read out of the coder's own transcript first, because "Claude
+# Code failed to implement the Work Item issue" names neither the cause nor
+# whether waiting would change it. When that reason turns out to be a spent
+# agent quota, retries hold until the reset the message names: every Work Item
+# draws on the same account, so retrying sooner only burns the budget.
+#
 # Prerequisites:
 #   - The harness is running (`ready-for-agent start`)
 #   - curl and jq on PATH
@@ -32,6 +41,8 @@
 #   scripts/drain-ready-issues.sh --repo owner/name --status
 #   scripts/drain-ready-issues.sh --repo owner/name --backend claude \
 #     --build-model sonnet --thinking-level high
+#   scripts/drain-ready-issues.sh --repo owner/name --max-retries 5
+#   scripts/drain-ready-issues.sh --repo owner/name --no-retry
 #
 # Honours READY_FOR_AGENT_GRAPHQL_URL for non-default ports, same as
 # `ready-for-agent add`.
@@ -45,10 +56,12 @@ THINKING_LEVEL=""
 AUTO_MERGE=""
 DRY_RUN=""
 STATUS_ONLY=""
+NO_RETRY=""
 POLL_SECONDS="${POLL_SECONDS:-60}"
+MAX_RETRIES="${MAX_RETRIES:-3}"
 
 usage() {
-  sed -n '2,36p' "$0" | sed 's|^# \{0,1\}||'
+  sed -n '2,47p' "$0" | sed 's|^# \{0,1\}||'
   exit "${1:-0}"
 }
 
@@ -59,6 +72,8 @@ while [[ $# -gt 0 ]]; do
     --build-model) BUILD_MODEL="${2:-}"; shift 2 ;;
     --thinking-level) THINKING_LEVEL="${2:-}"; shift 2 ;;
     --poll) POLL_SECONDS="${2:-}"; shift 2 ;;
+    --max-retries) MAX_RETRIES="${2:-}"; shift 2 ;;
+    --no-retry) NO_RETRY=1; shift ;;
     --auto-merge) AUTO_MERGE=1; shift ;;
     --dry-run) DRY_RUN=1; shift ;;
     --status) STATUS_ONLY=1; shift ;;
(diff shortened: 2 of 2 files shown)


cgbarlow:
Pushed a fourth commit, `retry failed Work Items and report why they failed`. It came out of running the script for real rather than from reading it again, so the description of the problem below is what actually happened on this box.

## What went wrong

I admitted 10 Issues and walked away. Six Work Items reached Build and all six failed inside thirteen seconds of each other, with the same line:

```
#197 Build / Failed  Claude Code failed to implement the Work Item issue
```

Two things were wrong, and only one of them was mine.

**Mine:** a failed build is not terminal. The harness parks the Work Item with `canRetry` set and waits for someone to press Retry. My follow loop treated "not terminal" as "still moving", so it polled every 60s, forever, over ten Work Items that could never move — four of them queued behind blockers that would never close. For the unattended case this PR exists to serve, that is the worst available shape: it looks busy while doing nothing, indefinitely.

**Not mine, but worth your eye:** the reason was recorded and not shown. Every one of those six transcripts ended:

```
You've hit your session limit · resets 4:20am (UTC)
```

`statusMessage` said only that Claude Code failed. That names neither the cause nor — the part that decides what to do next — whether waiting would change it. It reads like a code failure and it was a quota wall. This is the same shape as points 1, 4 and 5 in #933: the harness knows, and doesn't say.

## What the commit does

- **Presses Retry**, bounded by `--max-retries` (default 3). A Work Item that fails identically three times is not waiting on patience.
- **Reads the reason from the coder's own transcript** and prints it beside the Work Item. This is the part I'd most like a second opinion on: it reaches into `~/.claude/projects` for Claude Code and returns nothing for every other backend, which is honest but lopsided. If the harness surfaced the agent's last message on a failed Step, the script wouldn't need to — and the UI would be better for it too.
- **Treats a spent quota as one wall, not N failures.** Every Work Item on the machine draws on the same account, so retrying before the reset spends the retry budget on a wall that hasn't moved. It parses the reset time and holds until then.
- **Records the wall rather than sleeping through it.** My first attempt slept in place; the loop stopped following the Work Items still building, and those were exactly the ones about to hit the same wall. It now keeps polling and simply withholds retries.
- **Stops instead of spinning** when nothing can move by itself — every unfinished Work Item has failed or is queued behind one that has — and returns non-zero.
- `--status` stays read-only: it reports the reasons and presses nothing.

## Verified

Retrying moved all six Work Items off Failed and two reached Review. One hit the wall a second time; the hold parsed the new reset time, held 4h23m, and resumed on its own. That path is now tested by having happened rather than by being simulated.

One observation you may find more useful than the patch. Six concurrent builds on Opus at high Thinking Level exhausted a session quota in roughly fifty seconds of real work, with `maxConcurrentWorkItems` at 5 and `maxConcurrentAgentTurns` at 2 — the shipped defaults. An unattended drain is the workload most likely to find that ceiling and least likely to have anyone around when it does. A note in the docs that concurrency multiplies quota burn, or a warning when a failure is quota-shaped, would land well before anything in this script does.

cgbarlow:
Pushed a fix for a bug in the retry code I added, found by leaving the drain running overnight rather than by re-reading it.

The quota hold compared `resets 2:20pm` against the current time. Inside the same hour that reads correctly, which is why it tested green and why the 4h23m hold in the description genuinely worked. Once the drain outlived its own reset the comparison inverted: the clock now sat in the past, the code inferred tomorrow, and the loop idled 5h20m on a quota that had come back at 14:20. Failures were at 09:35; the reset was the same day. It now reads the clock against the failure's own timestamp and holds nothing once the reset has passed.

Carrying that timestamp exposed a second one. `retry_failures` reads the reason via `reason="$(backend_failure_reason ...)"`, so my first attempt — set a global inside that function — assigned it in a subshell and the caller never saw it. It would have been empty on every call and silently fallen back to the current time, i.e. the original bug with a fix on top of it. Both values now come back together on stdout.

Worth flagging for the harness itself: nothing here is specific to my script. Any consumer parsing that reset notice has the same anchoring problem, and the notice is only reachable by reading the coder's transcript in the first place. A failed Step carrying its agent's last message, and a machine-readable resets-at, would remove the need for both the transcript reader and this parser.

Also updated the description: a Work Item has now been observed through to a merged PR under `--auto-merge`, so that caveat is gone. The unhappy path — a PR failing its checks, and `waitForReadyForReviewChecks` holding it — is still unobserved.

berenddeboer:
@cgbarlow I feel this belongs more inside the harness, and not a separate script. The contract and life-cycle are better done in the harness. The key thing I believe you try to accomplish is the ability to queue things so they start at a given time? I.e. the deal with limited subscription quota?

PR #952, opened 2026-08-11
fix(github-service): diagnose corporate TLS inspection instead of retrying as GitHub failures

### LINKED ISSUE #949: Corporate TLS inspection breaks every GitHub API call, reported as "Failed to list Ready-labeled Issues" / "Failed to look up an open pull request"
Running v0.19.0 on macOS 26.6 (arm64) on a corporate laptop with Netskope TLS
inspection, the Harness looked alive but silently did no GitHub work for 45
minutes. Every GitHub API call failed; every message named the wrong culprit.

## Symptoms

```
[09:43:32.042] ERROR (#51): Scheduled Issue poll failed {
  jobId: "qjob-01KZNB6AGH12KEHG2B2KMNC0R4",
  repositoryId: "repo-01KZNB6AGEN2QMFP56HK8EC4TM",
  error: "Failed to list Ready-labeled Issues for <owner>/<repo>",
}
```

and, on the same Repository:

```
step_run.reason_message =
  "Failed to look up an open pull request for <owner>/<repo>:rfa/<slug>/1/wi-01KZ..."
```

Both point at GitHub. Neither is a GitHub problem.

## Actual cause

The corporate proxy MITMs `api.github.com`:

```
$ echo | openssl s_client -connect api.github.com:443 -servername api.github.com
 0 s:/CN=*.github.com
   i:/CN=ca.deloitte-apa-wps.au.goskope.com/O=... /emailAddress=certadmin@netskope.com
 2 s:/C=AU/O=Netskope Inc./CN=certadmin      <- self-signed root
```

The Netskope root is trusted in the macOS System keychain, so `git`, `gh` and
`curl` all work — which is exactly why this is so hard to spot. Bun does not
read the keychain, so every `fetch` in the Harness fails:

```
FETCH FAILED: TypeError fetch failed  cause: SELF_SIGNED_CERT_IN_CHAIN
```

Confirmed by running the same GraphQL query two ways with the same
`gh auth token` value: bare `fetch` fails with `SELF_SIGNED_CERT_IN_CHAIN`,
`NODE_EXTRA_CA_CERTS=<netskope-root.pem> fetch` returns `HTTP 200`.

Workaround, for anyone who lands here:

```sh
security find-certificate -a -c certadmin -p /Library/Keychains/System.keychain > ~/.config/netskope-ca.pem
export NODE_EXTRA_CA_CERTS=~/.config/netskope-ca.pem
```

### 1. A permanent TLS failure is classified as retryable

`packages/github-service/src/lib/github-service-live.ts:124-134` — the `catch`
in `githubRequest` only special-cases `GitHubHttpError` (retryable on 5xx) and
`GenqlError` (never retryable). Everything else, including a TLS handshake
rejection, falls into the `{ retryable: true }` default. `githubQuery`
(`:150-162`) then retries 3x/500ms per call.

`SELF_SIGNED_CERT_IN_CHAIN`, `UNABLE_TO_VERIFY_LEAF_SIGNATURE`,
`DEPTH_ZERO_SELF_SIGNED_CERT`, `CERT_HAS_EXPIRED` and friends are all
permanent for the process lifetime. Retrying them wastes three round trips and,
worse, dresses a configuration error up as a transient one. These should be
classified non-retryable and carry a distinct error identity rather than being
flattened into a generic `GitHubRequestError`.

### 2. Nothing tells the operator TLS is the problem

There is no preflight. The failure only manifests per-operation, wrapped in
per-operation prose, and the `cause` is dropped before display (see the
companion issue on error causes). The Harness has all the information at
startup and never uses it.

## Suggested fix

- Classify TLS/certificate `cause.code` values as non-retryable in
  `githubRequest`, and surface them as their own tagged error
  (e.g. `GitHubTlsTrustError`) so callers and log formatting can name them.
- Add a startup preflight that makes one cheap authenticated request per
  configured forge host. On a TLS trust failure, fail loudly and name the
  remedy, e.g.

  ```
  Cannot establish a trusted TLS connection to api.github.com.
  The certificate chain is signed by "Netskope Inc." rather than a public CA,
  which usually means a corporate TLS-inspection proxy. git/gh/curl trust it
  via the OS trust store, but Ready for Agent does not read that store.
  Export the root CA and point NODE_EXTRA_CA_CERTS at it:
    security find-certificate -a -c certadmin -p /Library/Keychains/System.keychain > ~/.config/corp-ca.pem
    export NODE_EXTRA_CA_CERTS=~/.config/corp-ca.pem
  ```

  Pulling the issuer O/CN out of the presented chain makes this
  self-diagnosing, and generalises to Zscaler / Palo Alto / mitmproxy.
- Document `NODE_EXTRA_CA_CERTS` in the README/troubleshooting docs. Bun
  honours it, so it works for the compiled binary.

## Collateral damage worth noting

Because `git` over SSH kept working while the API did not, the Harness looked
functional and wedged a Work Item halfway:

- `repository.issues_reconciled_at` stayed at `09:04:29` while the process
  started at `09:07` — this run never completed a poll, and the UI happily
  served Issue rows written the previous evening. A poll that fails forever
  reschedules cleanly (`apps/harness/src/server/job-worker.ts:340-348`) and is
  invisible unless you read the log.
- `commit` succeeded at 09:11:45, then `create_pr` failed at 09:11:47 on the
  first API call of the Step (`packages/work-item-lifecycle/src/lib/create-pr.ts:772`),
  which is before the push — so the branch never reached the remote
  (`git ls-remote` showed only `main`) and the `jobs` queue entry burned
  through `attempts=2/2` with no path back.

Separately from the TLS fix: a Repository whose polling has been failing for N
consecutive attempts should say so in the UI rather than showing stale Issues
as though they were current.

berenddeboer:
## Why

Corporate TLS-inspection proxies (e.g. Netskope) present a private root CA
that git/gh/curl trust via the OS keychain, but Bun does not. Every GitHub
API fetch failed with SELF_SIGNED_CERT_IN_CHAIN while the harness looked
healthy: poll and Step errors named GitHub, permanent cert failures were
treated as retryable, and there was no startup signal or NODE_EXTRA_CA_CERTS
guidance.

## What changed

- Classify TLS OpenSSL codes as non-retryable GitHubTlsTrustError (no query
  retries).
- Startup forge TLS preflight probes api.github.com and each GitLab forge
  host; fail loudly with issuer when readable and remediation.
- Keymaxxer: helper exit code 5 + non-secret github-tls-trust control; parent
  rebuilds operator remediation.
- Count-lite returns tls-trust; reconciler and lifecycle error unions updated.
- Document NODE_EXTRA_CA_CERTS in README troubleshooting.

## Verification

- github-service unit tests (TLS classification, non-retry, protocol, count)
- github-service typecheck
- ready-for-agent forge TLS / peek-endpoint unit tests
- Pre-commit affected checks (including issue-reconciler error channel)

## Limitations

- Preflight uses an unauthenticated HTTPS probe (TLS handshake only).
- Remediation copy is MITM / NODE_EXTRA_CA_CERTS-centric for all TLS codes.
- Lifecycle step messages keep the 500-char cap; full remedy on preflight and
  uncapped poll logs.

Closes #949

### DIFF
--- a/README.md
+++ b/README.md
@@ -239,6 +239,34 @@ repositories (`gh` for GitHub, `glab` for GitLab) block startup. A
 missing coding agent never does — it shows as Unavailable instead
 (see below).
 
+### Startup fails with "Cannot establish a trusted TLS connection"
+
+Corporate TLS-inspection proxies (Netskope, Zscaler, Palo Alto,
+mitmproxy, and similar) present a private root CA. `git`, `gh`, and
+`curl` usually trust it via the OS keychain, but Ready for Agent runs
+on Bun, which does **not** read that store — every GitHub/GitLab API
+call then fails with a certificate error such as
+`SELF_SIGNED_CERT_IN_CHAIN`.
+
+On startup the harness probes each configured forge API host. When
+TLS trust fails it stops immediately and prints the remedy. Export
+your corporate root CA and point Bun at it with
+`NODE_EXTRA_CA_CERTS` (honoured by the compiled binary as well):
+
+```bash
+# macOS — Netskope example (adjust -c to your proxy CA common name):
+security find-certificate -a -c certadmin -p \
+  /Library/Keychains/System.keychain > ~/.config/corp-ca.pem
+export NODE_EXTRA_CA_CERTS=~/.config/corp-ca.pem
+
+# Linux — use the PEM your IT provides:
+export NODE_EXTRA_CA_CERTS=/path/to/corp-root-ca.pem
+```
+
+Set the variable in your shell profile (or the service unit that
+starts the harness) so it applies on every launch, then restart
+`ready-for-agent`.
+
 ### A labelled issue does not show up
 
 - The issue must carry the `ready-for-agent` label — the harness only
--- a/apps/harness/src/server/ambient-github-layer.ts
+++ b/apps/harness/src/server/ambient-github-layer.ts
@@ -2,9 +2,9 @@ import { Cache, Duration, Effect, Exit, Fiber, Layer, Option } from "effect"
 import { ChildProcess, ChildProcessSpawner } from "effect/unstable/process"
 import {
   type GitHubOperationOptions,
-  type GitHubRepositoryUnavailableError,
   GitHubRequestError,
   GitHubService,
+  type GitHubServiceError,
   type GitHubServiceShape,
   type GitHubThrottledError,
   isGitHubThrottledError,
@@ -14,11 +14,6 @@ import {
   GitHubOperationCoordinator,
   type GitHubOperationOrigin,
 } from "./github-operation-coordinator.js"
-
-type GitHubServiceError =
-  | GitHubRepositoryUnavailableError
-  | GitHubRequestError
-  | GitHubThrottledError
 
 /** Unit key for the process-wide ambient GitHub CLI token cache. */
 const TOKEN_CACHE_KEY = true as const
--- a/apps/harness/src/server/keymaxxer-github-layer.ts
+++ b/apps/harness/src/server/keymaxxer-github-layer.ts
@@ -11,15 +11,19 @@ import {
 import {
   GITHUB_HELPER_AUTHENTICATION_EXIT_CODE,
   GITHUB_HELPER_THROTTLED_EXIT_CODE,
+  GITHUB_HELPER_TLS_TRUST_EXIT_CODE,
   type GitHubHelperOperation,
   type GitHubOperationOptions,
   type GitHubRepository,
   GitHubRepositoryUnavailableError,
   GitHubRequestError,
   GitHubService,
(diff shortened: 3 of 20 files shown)


berenddeboer:
### Status check fix `fd476ecd6f705b93e058b17f8803c4ad93c15eab`

**Cause:** `quality-gates` failed on `ready-for-agent:knip` — unused exports:
- `readPresentedCertificateIssuer`
- type `ForgeTlsPreflightFetch`

**Change:** keep both module-private (no public export). Behavior of forge TLS preflight is unchanged.

**Verification:**
- `bunx nx run ready-for-agent:knip`
- `bun test` forge-tls-preflight tests
- pre-commit nx-affected lint/typecheck/test

**Automated review:** none present on this PR (no review comments/reviews to address). Feedback declined: none.

PR #959, opened 2026-08-11
fix(harness): surface stale issue projections and back off failed polls

### LINKED ISSUE #951: A Repository whose scheduled Issue poll fails forever looks healthy: stale Issues served as current, no UI signal
Split out of #949, which was a TLS misconfiguration. This issue is about the
part that was not TLS-specific: **a Repository whose scheduled Issue poll fails
indefinitely is indistinguishable, in the UI, from one that is healthy.**

Any persistent poll failure has this shape — expired token, revoked scope,
repository renamed or access removed, DNS, proxy, forge outage. The TLS fix
removes one cause, not the blind spot.

## What I observed

For 45 minutes the Harness showed a Repository with its Issues listed and
nothing else. Every scheduled poll in that window failed. The Issue rows on
screen had been written the previous evening:

```
repository.issues_reconciled_at  = 2026-08-11 09:04:29   (previous process)
harness process started          = 2026-08-11 09:07
wall clock                       = 2026-08-11 09:51
issue.updated_at                 = 2026-08-10 20:33
```

So this process had never completed a poll, and the UI served 13-hour-old data
as current. The only evidence was one `ERROR` line per minute in the terminal.
I did not go looking for it, because nothing on screen suggested anything was
wrong.

## Why it is invisible

**The failure is swallowed into a normal reschedule.**
`apps/harness/src/server/job-worker.ts:340-348` logs the error, then falls
through to `queue.postponeKeyed(jobId, delay)` at the ordinary cadence. No
attempt counter, no backoff, no state written anywhere. The `job_queue` row sat
at `job_attempts = 0` throughout — from the queue's perspective nothing had
ever failed.

**`issuesReconciledAt` is the only record of success, and is read as a
boolean.** It is plumbed all the way to the client
(`packages/graphql-api/src/lib/graphql-api.ts:760-762`, `schema.graphql:357`),
but both consumers only ask "is it null?":

- `apps/harness/src/home-page-content.tsx:2577` — `null` renders
  "Not refreshed yet."; anything non-null renders the Issue list, with no
  indication of age.
- `:1549-1552` — non-null enables the Issues query.

A timestamp from 13 hours ago and one from 10 seconds ago are treated
identically. The value needed to detect staleness is already on the client and
simply isn't compared to the clock.

**The one error banner is unrelated.** `home-page-content.tsx:2570-2575`
("Failed to refresh issues.") is bound to the *manual* refresh mutation. A
failing *scheduled* poll never surfaces it.

**This is by design, and the design is what I'm questioning.**
`ARCHITECTURE.md:14`:

> Successful Issue reconciliation publishes a Repository-specific issues
> invalidation and a repositories invalidation so subscribed clients refetch
> Issues and `issuesReconciledAt`; no client-facing job history or status
> snapshots are maintained

Only *success* is a publishable event, and status snapshots are explicitly out
of scope. That keeps the model clean, and I can see the appeal, but it means
failure is unrepresentable — the UI cannot render a fact the architecture
refuses to store. So the fix is a deliberate decision, not an oversight to
patch.

## Options, cheapest first

1. **Client-side staleness only, no schema change.** The client already has
   `issuesReconciledAt`. If it is older than a small multiple of the poll
   interval, mark the Repository stale and caption the Issue list with
   "Last refreshed 13 hours ago". Costs nothing, needs no new state, and would
   have caught this. Doesn't distinguish "failing" from "Harness not running",
   which is fine — both mean "don't trust these Issues".
2. **Consecutive-failure count on the Repository.** Reset on success, increment
   in the `job-worker.ts:340` branch. Above a threshold, show a banner naming
   the last error (needs #950 to make the error worth showing). This is the
   smallest thing that admits failure into the model.
3. **Back off on repeated failure.** Independent of any UI: retrying a
   permanent failure at full cadence forever is wasteful, and with #949 it was
   3 retries per call inside each poll. Exponential backoff to a ceiling would
   also make the log readable.

I'd take 1 and 3 regardless; 2 is the one that needs a call on the
ARCHITECTURE.md rule.

## Note

Ready for Agent runs unattended — that is the entire point. A silent, unbounded
"I have stopped looking at your Issues but the screen looks normal" is the worst
failure mode for that use case, and it costs a stale Issue list rather than a
visible error.

berenddeboer:
Why

When scheduled Issue reconciliation fails indefinitely (expired token, TLS,
renamed repo, forge outage, etc.), the UI still showed a healthy Repository
with the last successful Issue list. `issuesReconciledAt` was treated only as
null vs non-null, so multi-hour-old data looked current. Failures were only
logged, then rescheduled at the normal ~60–90s cadence, so permanent failures
spammed the log and the forge.

What changed

Client-side staleness (no schema change)
- Treat a non-null `issuesReconciledAt` older than 5x the poll upper bound
  (7.5 minutes) as a stale Issue projection.
- On the Repository card, show a guidance Stale banner with a relative caption
  (e.g. "Last refreshed 13 hours ago") and note that Issues may be out of date;
  still show the list so operators can see what is cached.
- Tick the wall clock so a mounted card can cross the threshold without a
  successful poll invalidation.

Scheduled-poll failure backoff
- Track consecutive failures per Repository in process memory (`postponeKeyed`
  resets queue `job_attempts`, so the queue cannot count this).
- On failure, postpone with exponential backoff from the healthy base
  (60s → … → 15m cap); log `consecutiveFailures` and `nextDelayMs`.
- On success, clear the counter and restore the normal sampled cadence.
  GitHub throttle postponement is unchanged and does not count toward failure
  backoff.

Intentionally not in this change: a durable consecutive-failure count or
last-error field on the Repository (that would need an ARCHITECTURE.md
decision).

Verification

- Unit tests for freshness helpers, issue-poll backoff math, job-worker
  failure/success delay paths, and repos interchange wiring for the Stale
  banner.
- `graphql-api` and `harness` typecheck; pre-commit lint/typecheck/test green
  after Biome format fixes.

Limitations

- Staleness is inferred from last-success age only — it does not distinguish
  "polls failing" from "Harness not running."
- Failure backoff counters are process-local and reset on restart.
- Stale UI is on the Repository Issues card; it does not add a new GraphQL
  field or job-status snapshot.

Closes #951

### DIFF
--- a/apps/harness/src/home-page-content.tsx
+++ b/apps/harness/src/home-page-content.tsx
@@ -51,6 +51,10 @@ import {
   issueActionEligibility,
 } from "./issue-action-eligibility.js"
 import {
+  formatLastRefreshedAgo,
+  isIssueProjectionStale,
+} from "./issue-projection-freshness.js"
+import {
   formatDuration,
   formatStartedAgo,
   isLiveDurationStatus,
@@ -1550,6 +1554,13 @@ function RepositoryCard({
   })
   const hasNoRelevantIssues =
     repository.issuesReconciledAt !== null && relevantIssues?.length === 0
+  // Wall-clock tick so a previously-fresh projection can cross the stale
+  // threshold while this card stays mounted (failed polls never invalidate).
+  const nowMs = useNowMs(repository.issuesReconciledAt !== null, 30_000)
+  const issuesProjectionStale = isIssueProjectionStale(
+    repository.issuesReconciledAt,
+    nowMs,
+  )
 
   const addGitHubToken = useMutation({
     mutationFn: async () => {
@@ -2577,13 +2588,29 @@ function RepositoryCard({
             {repository.issuesReconciledAt === null ? (
               <p className={ui.repoIssuesUnrefreshed}>Not refreshed yet.</p>
             ) : (
-              <Suspense fallback={<RepositoryIssuesSkeleton />}>
-                <RepositoryIssues
-                  repository={repository}
-                  workItems={workItems}
-                  workItemsLoading={workItemsLoading}
-                />
-              </Suspense>
+              <>
+                {issuesProjectionStale && (
+                  <Banner
+                    className={cx(ui.bannerCompact, "mb-2")}
+                    tone="guidance"
+                    tag="Stale"
+                    role="status"
+                  >
+                    {formatLastRefreshedAgo(
+                      repository.issuesReconciledAt,
+                      nowMs,
+                    )}
+                    . Issues may be out of date.
+                  </Banner>
+                )}
+                <Suspense fallback={<RepositoryIssuesSkeleton />}>
+                  <RepositoryIssues
+                    repository={repository}
+                    workItems={workItems}
+                    workItemsLoading={workItemsLoading}
+                  />
+                </Suspense>
+              </>
             )}
           </div>
           {removeRepository.isError && (
--- a/apps/harness/src/issue-projection-freshness.ts
+++ b/apps/harness/src/issue-projection-freshness.ts
@@ -1,0 +1,68 @@
+/**
+ * Client-side freshness for a Repository's Issue projection.
+ *
+ * Scheduled Issue polls run about every 60–90s
+ * (`ISSUE_POLLING_BASE_SECONDS` + jitter in `@ready-for-agent/graphql-api`).
+ * Only successful reconciliation advances `issuesReconciledAt`. A non-null
+ * timestamp that is older than a small multiple of that cadence means the
+ * Issues on screen must not be trusted as current — whether the Harness is
+ * failing polls, paused, or not running.
+ *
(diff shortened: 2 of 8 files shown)


berenddeboer:
## Status check handoff fix

**Commit:** `7393ba59d95fdb640a6c6374eff6b57174e03ce4`

### Changes
- Fixed `quality-gates` / `harness:knip` by unexporting unused module-private constants in `issue-projection-freshness.ts` (`ISSUE_POLL_CADENCE_UPPER_MS`, `ISSUE_PROJECTION_STALE_MULTIPLE`).
- `ISSUE_PROJECTION_STALE_AFTER_MS` remains exported for tests/callers.

### Verification
- `bunx nx run harness:knip` — pass
- `issue-projection-freshness` unit tests — pass
- Pre-commit `nx affected -t lint,typecheck,test` — pass

### Review feedback
- No automated review comments or completed reviewer feedback on this PR.
- None declined.

PR #972, opened 2026-08-12
docs(cli): define Repository Intake automation

berenddeboer:
## Why

PR #934 established the need to discover and start all currently eligible Repository work and inspect its progress through GraphQL against a running Harness. That behavior belongs in the published operator CLI, while lifecycle classification and Kanban meaning must remain authoritative inside the Harness rather than a repository-local shell script.

## What Changed

- define `Repository Intake` and `Intake Candidate` in the domain glossary and ontology
- record the Harness-owned GraphQL and versioned JSON architecture in ADR 0051
- specify `candidates`, `intake`, and `status` commands, including best-effort failure and exit semantics
- move Kanban lane ownership to one server projection shared by the visual board and CLI status
- explicitly exclude refresh, dry run, Intake IDs, concurrency flags, retries, settings mutations, and script compatibility

Related: #934


berenddeboer:
This will replace #934 .

PR #1049, opened 2026-08-14
fix(ready-for-agent): compile linux-x64 as bun-linux-x64-baseline

### LINKED ISSUE #1047: Linux x64 published binary SIGILLs on pre-Haswell CPUs (no AVX2/BMI2)
## Problem

`npx ready-for-agent@latest` (and `npx ready-for-agent@0.21.0` / `@0.22.0`) never starts the Harness on some Linux x64 hosts. The Node launcher installs `ready-for-agent-linux-x64` and execs it; that binary dies immediately with `SIGILL` (`Illegal instruction`, exit 132). Nothing binds `127.0.0.1:6056`.

`npx` often prints nothing — the crash is swallowed — so it looks like the CLI did not start.

Reproduced on Intel Core i7-3770 (Ivy Bridge, 2012): AVX present, **no AVX2, no BMI2**. gdb first fault:

```
=> 0x3cb6c0d:  shlx   %rdx,%rax,%rax
```

The same machine runs local Bun 1.3.14 and `bun run ready-for-agent` from a source checkout without issue.

## Root cause

`bunCompileTarget("linux-x64")` in `apps/ready-for-agent/bin/select-platform.js` returns `bun-linux-x64`.

Bun documents that the default x64 compile target uses AVX2/Haswell SIMD, and that `"Illegal instruction"` means use `bun-linux-x64-baseline` (Nehalem, pre-2013):

https://bun.com/docs/bundler/executables

The published 0.22.0 linux-x64 binary contains tens of thousands of AVX2-class instructions. The launcher (`bin/ready-for-agent.js`) is plain Node and runs on this CPU; it has no SIGILL handling, so the operator sees a silent failure.

This does **not** violate ADR 0023 (single launcher + platform `optionalDependencies`). The package shape is correct; only the Bun `--target` for linux-x64 is too new.

## Required behavior

- `npx ready-for-agent@latest` starts the Harness on Linux x64 CPUs that have SSE4.2 but not AVX2/BMI2 (Ivy Bridge and similar).
- The public package set stays the existing six platform packages. Do **not** add a seventh `*-baseline` optionalDependency (npm would install both on every linux-x64 host and roughly double the download).
- Do **not** add a postinstall GitHub download (rejected in ADR 0023).
- PR CI, `packed-install-smoke`, overnight published-install, and the six-platform release compile/publish/release-asset loop stay the same shape. Release still compiles the same platform keys; only the linux-x64 Bun target string changes.
- GitHub-hosted `ubuntu-latest` runners (Haswell+) must still run the packed-install and overnight smokes.
- If a binary still dies with SIGILL, the launcher must print a short message naming the CPU-feature mismatch and the baseline target — not exit 1 with empty output.

## Suggested fix (preferred)

One compile-target change, plus a launcher diagnostic:

1. `bunCompileTarget("linux-x64")` → `"bun-linux-x64-baseline"`.
2. Update the 1:1 assertion in `apps/ready-for-agent/src/select-platform.test.ts`.
3. In `apps/ready-for-agent/bin/ready-for-agent.js`, if `spawnSync` returns `signal === "SIGILL"` (or `status === null` after an illegal-instruction death), print a one-paragraph error that names `bun-linux-x64-baseline` / pre-Haswell CPUs.
4. Mention Linux x64 baseline in the product README troubleshooting (same section style as the TLS-inspection note).
5. Optional one-line note on ADR 0023: linux-x64 is the Bun baseline target so pre-Haswell hosts can run the published binary.

Do **not** introduce `ready-for-agent-linux-x64-baseline` as a seventh package. npm `os`/`cpu` cannot express AVX2, so both packages would install on every linux-x64 machine.

## Acceptance criteria

- [ ] `bunCompileTarget("linux-x64")` is `bun-linux-x64-baseline`
- [ ] `select-platform.test.ts` asserts that mapping; other five targets unchanged
- [ ] Release compile loop, `PLATFORM_PACKAGE_NAMES`, and optionalDeps still list exactly the current six packages
- [ ] `packed-install-smoke` and host-binary smoke still pass on `ubuntu-latest` (baseline binaries run on Haswell+)
- [ ] Launcher prints a SIGILL-specific error instead of silent exit (unit-testable on the spawn result; do not require an Ivy Bridge runner in CI)
- [ ] Product README troubleshooting mentions the SIGILL / baseline case
- [ ] No new CI job, no new publish package, no postinstall download

## Implementation notes

| Area | Path |
| --- | --- |
| Target map | `apps/ready-for-agent/bin/select-platform.js` (`bunCompileTarget`) |
| Tests | `apps/ready-for-agent/src/select-platform.test.ts` |
| Launcher | `apps/ready-for-agent/bin/ready-for-agent.js` |
| Compile (no loop change) | `apps/ready-for-agent/scripts/compile-platform-binary.ts` |
| ADR (optional sentence) | `docs/adr/0023-public-npm-binary-distribution.md` |
| README | repository root `README.md` troubleshooting |

## Out of scope

- Windows x64 baseline (`bun-windows-x64-baseline`) — same Bun warning, separate ticket if someone reproduces SIGILL there
- Darwin x64 baseline — Bun says this rarely matters
- Dual modern+baseline packages or CPU-feature install selection
- Changing ADR 0023’s launcher + optionalDependency shape

## Evidence

- Host: Intel Core i7-3770, Linux 6.8, glibc 2.39, Node 22
- `/proc/cpuinfo` flags include `avx`, not `avx2` or `bmi2`
- Published `ready-for-agent-linux-x64@0.21.0` and `@0.22.0` both SIGILL on `--help` / `start`
- Source path `bun run ready-for-agent -- --no-open` serves UI 200 and GraphQL `{ __typename }` at `127.0.0.1:6056`

dieuwedeboer:
## Why

`npx ready-for-agent@latest` never starts the Harness on pre-Haswell Linux x64 CPUs. The Node launcher is fine; the published `ready-for-agent-linux-x64` binary is compiled with Bun's default Haswell/AVX2 target and dies immediately with `SIGILL`. Reproduced on an Intel Core i7-3770 (AVX, no AVX2/BMI2); gdb first fault is `shlx`. `npx` swallows the crash, so it looks like the CLI did nothing.

## What

Preferred fix from #1047 — change the compile target, not the package matrix:

- `bunCompileTarget("linux-x64")` is now `bun-linux-x64-baseline`.
- The other five platform targets are unchanged.
- `PLATFORM_PACKAGE_NAMES` / optionalDependencies stay the existing six packages (no seventh `*-baseline` package — npm `os`/`cpu` cannot express AVX2).
- The launcher reports `SIGILL` via a pure `binarySpawnFailureMessage` seam instead of exiting 1 with empty output.
- Product README troubleshooting and ADR 0023 note the baseline target.

## Verification

- `bun --conditions=@ready-for-agent/source test src/select-platform.test.ts` — 10 pass, including the baseline mapping, SIGILL copy, spawn-error preference, and launcher import guard.
- `bunx biome check` on the touched JS/TS files — clean.

Packed-install / host-binary smokes are unchanged in shape and still belong on Haswell+ `ubuntu-latest` (baseline binaries run there). They were not re-run in this checkout.

Closes #1047


berenddeboer:
Review against issue #1047 and the owner clarification in https://github.com/berenddeboer/ready-for-agent/issues/1047#issuecomment-5287908277:

1. **SIGILL diagnosis needs to be scoped to Linux x64.** `binarySpawnFailureMessage` receives only the spawn result, so any `SIGILL` from Linux ARM64, macOS, or another cause is reported as a Linux x64 AVX2/BMI2 mismatch. The message also states that the CPU lacks AVX2/BMI2 as fact, although the new baseline binary intentionally does not require those features; a SIGILL from the baseline binary could instead mean missing SSE4.2 or an unrelated illegal instruction. Please pass the selected platform into this decision, restrict the baseline-specific diagnosis to `linux-x64`, and describe the cause as likely rather than certain. (`apps/ready-for-agent/bin/select-platform.js:153-177`, `apps/ready-for-agent/bin/ready-for-agent.js:62`)

2. **Document the actual SSE4.2 compatibility floor.** `README.md` and ADR 0023 say broadly that pre-Haswell CPUs can run the baseline binary. The issue requirement is narrower: CPUs with SSE4.2 but without AVX2/BMI2. Older x64 CPUs without SSE4.2 remain unsupported. (`README.md:265-267`, `docs/adr/0023-public-npm-binary-distribution.md:3`)

3. **The launcher behavior is not exercised end to end.** The new test verifies the pure helper, while the launcher assertion only checks that the helper name occurs in the source. It does not prove that a signaled child produces the expected stderr and exit status. The packed-install smoke acceptance criterion was also not rerun, as noted in the PR body. (`apps/ready-for-agent/src/select-platform.test.ts:154-198`)

Everything else matches the requested shape: `linux-x64` maps to `bun-linux-x64-baseline`, the other five targets and six-package matrix are unchanged, no postinstall download or extra package was added, the focused test passes 10/10, and all current PR checks are green.

Issue #1066, opened 2026-08-14
Make AgentBackendExitError.message required via expand–contract


berenddeboer:
## What to build

Flip `AgentBackendExitError.message` from optional to required.

Issue #1056 added the field as **optional only as a staging decision**. A messageless exit error is not a legitimate constructed value — optionality existed so the diagnostics work did not have to rewrite every lifecycle stub in one review. This ticket is the contract half of that expand–contract.

## Current prerequisite status

- #1056 / #1067 completed the expand step: the optional, sanitized message exists and parsed adapter reasons reach it.
- #1057 / #1069 completed bounded stderr capture for Agent Turns.
- #1071 tracks the remaining readiness-inspection path, where stderr can still contain the only explanation and be discarded.

The schema must not become required until #1071 lands. Otherwise the type would claim every inspection exit has a meaningful message while the shared inspection runner can still throw the only available text away.

A truly silent non-zero process remains possible even after stderr capture. Handling that case belongs in this contract ticket: shared runners must synthesize a stable, non-empty fallback that identifies the backend failure and exit code rather than preserving optionality.

## Expand–contract

1. **Expand** — done in #1056: optional `message`, sanitized on construction, populated from parsed adapter and captured CLI output.
2. **Capture available diagnostics** — Agent Turns are done in #1057; readiness inspection is #1071.
3. **Migrate production** — every production construction supplies either the best available sanitized reason or a stable generic fallback for a silent exit.
4. **Migrate tests** — update stubs that still omit `message` while the field remains optional.
5. **Contract** — change `Schema.optionalKey(Schema.String)` to required `Schema.String` and delete the transitional doc comment.

## Acceptance criteria

- [ ] #1071 is closed before the schema contract changes.
- [ ] Every production `AgentBackendExitError` construction supplies a non-empty `message`.
- [ ] Parsed adapter reasons and captured sanitized stderr remain preferred over any generic fallback.
- [ ] A silent non-zero Agent Turn or readiness probe receives a stable non-empty fallback that includes the backend identity and exit code.
- [ ] Test stubs that construct the error also supply a message.
- [ ] The schema field is required (`Schema.String`), not `Schema.optionalKey`.
- [ ] The transitional optionality comment is removed.
- [ ] Tests cover silent exits for both turn and readiness/capture runners so future adapters cannot reintroduce a bare exit error.

## Out of scope

- Credential classification and lifecycle remediation from #1058.
[...]

Issue #1057, opened 2026-08-14
Capture a bounded stderr tail for Agent Turns


berenddeboer:
## What to build

When an Agent Backend CLI dies having written its only explanation to stderr, the operator sees that explanation.

**This is the fix for the originating incident.** With an expired AWS SSO token, Claude Code exited non-zero in about 2 seconds. Because the process exited, the startup-inactivity watchdog (`AgentBackendStartupTimeoutError`, 60s) never fired, and no JSONL `result` event was ever emitted — so the AWS credential error existed **only on stderr**, where it was discarded at the OS level. The operator saw `Claude Code fallback failed to install dependencies` and a cause chain ending in a bare `AgentBackendExitError`.

The preceding ticket alone would **not** have diagnosed this, because there was no parsed reason to preserve.

## Why stderr is ignored today, and why piping is still safe

`runCliTurn` hard-codes `captureStderr: false`, so turns spawn with `stderr: "ignore"`. Two comments defend it:

> Turns always ignore stderr so a chatty CLI cannot fill an undrained pipe.

> Never pipe stderr for turns (undrained stderr can deadlock).

That reasoning is sound but it argues against **not draining** a piped stream — not against piping. Two existing precedents in this codebase drain safely:

- The turn path already drains stdout and stderr concurrently at `concurrency: 2` before reading the exit code, in its capture helper.
- The Install Dependencies Step already does exactly the thing this ticket needs for a direct install command: pipe stderr, fold it into a bounded ring buffer that keeps only the last N characters, and run that fold concurrently with the exit-code wait. Its existing bound is 4 000 characters.

So the invariant to preserve is *"never leave a piped stream undrained"*, and it is preserved by construction if the fold is always part of the raced structure.

## Risks that must be covered by tests

These are the reasons the current code chose the blunt instrument, so each needs to be demonstrably handled rather than assumed:

- **Deadlock and unbounded memory.** A CLI writing far more than the bound to stderr must still complete, with retained memory bounded by the tail size — not by what the CLI produced.
- **Early tree-kill.** `runCliTurn` deliberately kills the process tree as soon as the finalize text arrives on stdout, without waiting for a natural exit. The stderr fold must sit inside the raced structure so it can neither leak a fiber nor delay or block that kill.
- **Interaction with the startup watchdog.** The watchdog is disarmed by the first stdout byte. Piping stderr must not change when it arms or disarms — a CLI that writes only to stderr should still hit the watchdog, not hang.
[...]

PR #1067, opened 2026-08-14
fix(agent-backend): surface parsed CLI exit reasons in cause chains

### LINKED ISSUE #1056: Carry Agent Backend exit failure reasons into operator-visible diagnostics
## What to build

When an Agent Backend CLI fails — an Agent Turn or a readiness probe — the operator sees **why**, not merely that it failed.

Today a failed Agent Turn surfaces in a Work Item's Cause chain disclosure as a bare `AgentBackendExitError` with no text and no exit code, **even when the backend adapter already parsed a human-readable reason and threw it away**. After this ticket the same failure reads:

```
InstallDependenciesFallbackError — Claude Code fallback failed to install dependencies
AgentBackendExitError 1 — <the reason the backend actually reported>
```

This is a pure diagnostics change. No behaviour changes, no new failure classification, no UI work.

## Background — where the text is lost

Four independent losses, verified against `794933d`:

1. **`AgentBackendExitError` has nowhere to put a reason.** It carries `exitCode`, `cwd`, optional `sessionId` and optional `classification` — no message and no output field.
2. **Adapters parse a reason and drop it.**
   - Claude: an `is_error` result is logged via `Effect.logWarning("Claude Code turn result is_error", { message })` and then a bare synthetic `exitCode: 1` is raised. The message never leaves the log.
   - Codex: the `turn.failed` reason is logged only. The code carries an explicit comment stating the reason is kept in logs *because* the exit error has no message field.
   - Grok: the stream `errorMessage` and the `maxTurnsReached` case are dropped **entirely** — not even logged.
   - `runCliCapture` holds captured stdout and stderr and discards both when it raises the exit error.
3. **The exit code never reaches the cause chain.** `linkFromValue` in the cause-chain extractor derives each link's `code` from `record.code` only. The error names its field `exitCode`, so nothing matches — which is exactly why the link renders with nothing after the tag.
4. **Two hand-rolled workarounds already exist for the missing field**, and are the strongest evidence it belongs there:
   - Codex's inspect path composes probe text manually, with a comment saying the exit error has no message for `formatInspectFailure`.
   - `formatInspectFailure` otherwise degrades to the useless string `"Agent Backend inspection failed (AgentBackendExitError)"`, which is what OpenCode and Grok inspect failures produce today.

## Design decisions already made

- **One `message` field, not `message` + `stderrTail`.** Two fields invite "which one do I read?" at every call site. Adapters compose the sentence; the field means "the best available human-readable reason".
- **Sanitize on the way in**, through the existing `sanitizeUserFacingText`, which strips ANSI escapes and bounds length. This value is persisted to the database and rendered in a browser.
- **No UI change and no GraphQL change.** The disclosure already renders `name code — message` per link, and the persisted cause chain is already exposed to operators (#1037). The presentation layer is correct; only the data feeding it is starved.
- **No new field on the cause-chain link type.** Deriving `code` from `exitCode` reuses the existing numeric-code handling.

## `message` is optional as a staging decision — it should ideally be required

**Optional is the compromise, not the intended end state.** Be explicit about this, because a future reader will otherwise read `optionalKey` as a deliberate statement that a messageless exit error is a legitimate thing to construct. It is not — a messageless exit error is precisely the defect this whole chain of work exists to remove.

The argument for **required**: an exit error with no reason is exactly the bug being fixed here. Making the field required means the type system refuses to let any adapter, present or future, raise a failure it cannot explain. That is the whole point — the compiler enforces the lesson instead of a comment asking future authors to remember it. With the field optional, the next backend adapter added to this repo can reintroduce the identical defect and nothing will stop it.

The argument for **optional now**: `AgentBackendExitError` is constructed in **23 places** — 6 in production (2 in the shared CLI runner, 1 in Claude, 1 in Codex, 2 in Grok) and 17 in tests across the agent-backend, opencode and work-item-lifecycle packages. *(Counted at `dd6649b`; re-count after pulling, as the ENOENT-classification work in `87789bb` may have added sites.)* Making it required in this ticket turns a focused diagnostics fix into a repo-wide edit that touches every lifecycle spec stub, which buries the change that matters under mechanical churn and makes the diff hard to review.

**So: ship optional here, and note in the schema's doc comment that optionality is transitional, with a pointer to the follow-up.** Do not describe it as a design preference.

Getting to required afterwards is a small expand–contract sequence, and the implementer should file it as a follow-up ticket when finishing this one:

1. *Expand* is already done by this ticket — the optional field exists.
2. *Migrate* the 6 production sites so every one supplies a message (this ticket already does 5 of them; the remaining runner site is covered once stderr capture lands).
3. *Migrate* the 17 test sites in batches by package, keeping CI green throughout because the field is still optional.
4. *Contract*: flip `optionalKey` to required once no caller omits it.

Note that step 4 cannot land before the stderr-capture ticket, because until then the shared CLI runner has one path with genuinely no text available to supply.

If, on starting the work, the 17 test sites turn out to be trivially mechanical, going straight to required in this ticket is the **better** outcome — take it. The optional staging exists to protect reviewability, not because the weaker type is preferable.

## Acceptance criteria

- [ ] `AgentBackendExitError` accepts a human-readable message, sanitized and length-bounded. If it is added as optional, its doc comment states that the optionality is transitional and points at the follow-up ticket.
- [ ] Claude populates it from the parsed `is_error` reason instead of only logging it.
- [ ] Codex populates it from the `turn.failed` reason, and its hand-rolled inspect-text workaround is deleted rather than left alongside the new field.
- [ ] Grok populates it from the stream `errorMessage` and from the `maxTurnsReached` case.
- [ ] `runCliCapture` populates it from output it has already captured.
- [ ] A cause-chain link derives its `code` from `exitCode` when no `code` field is present, so the exit code becomes visible.
- [ ] `formatInspectFailure` reports the backend's own message for an exit failure instead of `"Agent Backend inspection failed (AgentBackendExitError)"`.
- [ ] End to end: a Work Item whose Agent Turn fails with a backend-reported reason shows that reason **and** the exit code in its Cause chain disclosure.
- [ ] The Install Dependencies spec that currently asserts only `toBeInstanceOf(InstallDependenciesFallbackError)` is tightened to assert the backend's reason survives into the preserved cause.
- [ ] No secret or token-shaped text can reach the message: covered by a test.
- [ ] A follow-up ticket is filed to make the message required via expand–contract, unless this ticket already made it required.

## Explicitly out of scope

- Capturing stderr for Agent Turns — that is the next ticket, and it is the fix for the originating incident. This ticket only surfaces reasons the adapters already have.
- Recognising credential failures as a distinct class.

## Blocked by

None — can start immediately.

berenddeboer:
Operators were seeing a bare `AgentBackendExitError` with no text and no
exit code, even when Claude, Codex, or Grok had already parsed a
human-readable failure. Inspect copy degraded to "Agent Backend
inspection failed (AgentBackendExitError)". This is the diagnostics
half of #1056; stderr capture for turns remains #1057.

`AgentBackendExitError` now accepts an optional `message` (transitional
until #1066). It is sanitized on construction (ANSI stripped,
token-shaped secrets redacted, length bounded). Claude `is_error`,
Codex `turn.failed`, and Grok stream `error` / max-turns populate it.
`runCliCapture` uses captured stdout/stderr. Cause-chain links derive
`code` from `exitCode` so the disclosure can render
`AgentBackendExitError 1 — <reason>`. Codex inspect failures raise the
exit error with probe text instead of wrapping a ConfigError
workaround; `formatInspectFailure` already prefers `error.message`.

Review follow-up: a parsed stream reason now rides on `runCliTurn` the
same way `errorClassification` does, so a CLI that emits JSONL then
exits non-zero no longer drops the reason. Tests cover that path with
`exit 1` after the error event.

Verified with `agent-backend`, Claude, Codex, Grok, `github-service`
cause-chain, and install-dependencies fallback specs. Deferred:
duplicate sanitizer regexes stay package-local (layering); the nested
TypeScript-7 Nx symlink script is a local hook-layout fix, not product
behaviour.

Closes #1056

### DIFF
--- a/packages/agent-backend/src/lib/cli-runner.ts
+++ b/packages/agent-backend/src/lib/cli-runner.ts
@@ -25,6 +25,16 @@ import type { AgentBackendDescriptor, OnSessionId } from "./types.js"
 /** Graceful terminate then force-kill bound for the Agent Turn process tree. */
 export const DEFAULT_FORCE_KILL_AFTER = Duration.seconds(2)
 
+const capturedCliOutputMessage = (
+  stdout: string,
+  stderr: string,
+): string | undefined => {
+  const parts = [stdout, stderr]
+    .map((part) => part.trim())
+    .filter((part) => part.length > 0)
+  return parts.length > 0 ? parts.join("\n") : undefined
+}
+
 /**
  * Window from spawn to the first stdout output of an Agent Turn. A CLI that
  * crashes or hangs before emitting anything otherwise burns the whole turn
@@ -56,6 +66,13 @@ export type CliLineEvent = {
    * on to exit non-zero.
    */
   readonly errorClassification?: AgentBackendErrorClassification
+  /**
+   * Human-readable reason already parsed from the stream (e.g. Claude
+   * `is_error`, Codex `turn.failed`). Sticky like classification: once
+   * observed, it becomes AgentBackendExitError.message on a non-zero exit
+   * so the adapter does not have to run after the CLI dies.
+   */
+  readonly errorMessage?: string
 }
 
 export type RunCliCaptureInput = {
@@ -240,9 +257,11 @@ export const runCliCapture = (
     )
 
     if (result.exitCode !== 0 && input.allowNonZeroExit !== true) {
+      const message = capturedCliOutputMessage(result.stdout, result.stderr)
       return yield* new AgentBackendExitError({
         exitCode: result.exitCode,
         cwd: input.cwd,
+        ...(message !== undefined ? { message } : {}),
       })
     }
 
@@ -345,6 +364,7 @@ export const runCliTurn = (
               assistantText: string
               finalized: boolean
               errorClassification?: AgentBackendErrorClassification
+              errorMessage?: string
             } => ({
               assistantText: "",
               finalized: false,
@@ -359,6 +379,7 @@ export const runCliTurn = (
                 const sessionId = event.sessionId ?? acc.sessionId
                 const errorClassification =
                   event.errorClassification ?? acc.errorClassification
+                const errorMessage = event.errorMessage ?? acc.errorMessage
                 if (event.sessionId !== undefined) {
                   yield* Ref.set(seenSessionId, event.sessionId)
                   const alreadyNotified = yield* Ref.getAndSet(
@@ -400,6 +421,7 @@ export const runCliTurn = (
                     ...(errorClassification !== undefined
                       ? { errorClassification }
                       : {}),
+                    ...(errorMessage !== undefined ? { errorMessage } : {}),
                   }
                 }
 
@@ -415,6 +437,7 @@ export const runCliTurn = (
                   ...(errorClassification !== undefined
                     ? { errorClassification }
                     : {}),
+                  ...(errorMessage !== undefined ? { errorMessage } : {}),
                 }
               }),
           ),
@@ -435,6 +458,7 @@ export const runCliTurn = (
           assistantText: output.assistantText,
           finalized: output.finalized,
           errorClassification: output.errorClassification,
+          errorMessage: output.errorMessage,
         }
       }),
     ).pipe(
@@ -467,6 +491,9 @@ export const runCliTurn = (
           ...(sessionId !== undefined ? { sessionId } : {}),
           ...(result.errorClassification !== undefined
             ? { classification: result.errorClassification }
+            : {}),
+          ...(result.errorMessage !== undefined
+            ? { message: result.errorMessage }
             : {}),
         })
       }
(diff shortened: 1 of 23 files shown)


berenddeboer:
Pushed `2d064967dcbc2a0b7df7f2baa5e32db6dd042818` to fix the red `quality-gates` and `harness-e2e-live-forge` jobs.

**Change:** `AgentBackendExitError` no longer overrides the Schema constructor (TS377044 / `effect(overriddenSchemaConstructor)`). Sanitization now lives on static `AgentBackendExitError.new`, and call sites construct through that factory.

**Verification:** `nx run agent-backend:build`, `nx run agent-backend:typecheck`, and the exit-message / `runCliCapture` / `runCliTurn` specs passed. Pre-commit `nx affected -t lint,typecheck,test` passed on commit.

**Review feedback addressed:** none from Automated Review — this addresses the CI typecheck/build failure only.

berenddeboer:
Pushed `cb958a01e65eac062597f68af2a896296a5edd26` to fix the red `quality-gates` job.

**Change:** Stop exporting `AGENT_BACKEND_EXIT_MESSAGE_MAX`. The length bound stays private in the sanitizer; knip no longer reports an unused export.

**Verification:** `nx run agent-backend:knip` passed. Pre-commit `nx affected -t lint,typecheck,test` passed on commit.

**Review feedback addressed:** none from Automated Review — this addresses the CI knip failure only.

PR #1084, opened 2026-08-14
feat: reject Thinking Levels the governing Agent Model does not advertise

### LINKED ISSUE #1073: defaultThinkingLevel is stored without validating it against the selected model's levels
## Problem Statement

A Repository was able to persist this Agent Model selection:

```text
selectedAgentBackend: opencode
defaultModel:         opencode/deepseek-v4-flash
defaultThinkingLevel: medium
```

The selected model does not advertise `medium`. Its catalog entry reports
`thinkingLevels: ["low", "high", "max"]`. Other OpenCode models expose different
sets, so a Thinking Level cannot be validated independently of its Agent Model.

The browser derives Thinking Level choices from the selected catalog entry and clears
an incompatible draft when the operator actively changes a model. The server does not
enforce the same invariant. Its shared settings validator reduces the Agent Model
catalog to model IDs, discards each entry's `thinkingLevels`, and validates only
`defaultModel` and `reviewModel`. Both Harness Config and Repository settings can
therefore persist a valid model ID paired with an unsupported Thinking Level.

Repository settings are effectively whole-object writes for model preferences. A caller
that correctly carries the existing fields through an unrelated update can perpetuate
an invalid pair indefinitely.

Two details discovered during investigation correct the original report:

- Work Item enrollment currently validates model IDs only. An unsupported Thinking
  Level is not rejected there and can reach the Agent Backend CLI.
- A null Thinking Level does not always mean "Harness default." When an explicit
  Repository Agent Model has no Thinking Level, the runtime uses that model's
  backend/model default.

The failure should be attributed when an invalid pair is submitted and, for legacy data
or later catalog drift, before an Agent Backend CLI is spawned.

## Solution

Treat the advertised Thinking Levels as part of an Agent Model catalog entry's
executable contract.

- The browser continues to clear an incompatible draft when the operator actively
  changes a model.
- `updateConfig` and `updateRepositorySettings` reject an applicable non-null Thinking
  Level that the governing Agent Model does not advertise.
- The server never silently deletes or rewrites a persisted Thinking Level.
- The UI preserves an existing incompatible value as visibly unavailable, blocks Save,
  and gives the operator a direct way to choose a supported level or clear it.
- Work Item creation and pre-turn admission validate the complete resolved selection so
  legacy rows and catalog drift fail before an Agent Backend process starts.
- Persistence remains catalog-agnostic. Live catalog lookup and validation stay at the
  application/API and process-spawn boundaries.

This extends the catalog-only policy from issue #838 from Agent Model identity to the
per-model Thinking Level metadata already present in the same catalog.

## Decision: Reject Rather Than Silently Clear

The mutations must reject invalid submitted pairs rather than normalize them to null.

Why:

- Issue #838 deliberately preserves stale catalog values and requires an explicit
  operator correction. Silently clearing Thinking Levels would create a conflicting
  policy for another field from the same catalog.
- A catalog can change temporarily because of provider mode, account state, discovery,
  or an Agent Backend release. A server-side clear could destroy a preference that
  becomes valid again later.
- The mutations return only the updated Config or Repository. They have no warning or
  normalization-result channel through which to explain that the server changed the
  submitted value.
- Null has scope-dependent semantics. For an explicit model it means backend/model
  default, not necessarily the Harness Thinking Level.
- Rejection produces an existing structured field error and makes a direct GraphQL
  client correct its request rather than accepting a result different from the request.

The deliberate exception is the interactive browser draft. When the operator actively
chooses a different model, clearing a no-longer-applicable draft Thinking Level is an
immediate and visible consequence of that action and is existing behavior.

## Runtime Selection Semantics To Preserve

Validation must use the same governing Agent Model and fallback order as runtime model
resolution. It must not zip similarly named input fields without considering scope.

| Scope and role | Governing Agent Model | Thinking Level behavior |
| --- | --- | --- |
| Harness build | Harness `defaultModel` | Harness `defaultThinkingLevel`; null uses backend/model default |
| Harness review with explicit review model | Harness `reviewModel` | Harness `reviewThinkingLevel`; null uses that review model's backend/model default |
| Harness review without review model | Harness build model | Harness `reviewThinkingLevel`, then Harness build Thinking Level |
| Repository build with explicit build model | Repository `defaultModel` | Repository `defaultThinkingLevel`; null uses that model's backend/model default |
| Repository build without build model | Complete Harness build selection | Repository `defaultThinkingLevel` is dormant and does not override the inherited selection |
| Repository review with explicit review model | Repository `reviewModel` | Repository `reviewThinkingLevel`; null uses that model's backend/model default |
| Repository review without a Repository review model, with a Harness review model | Complete Harness review selection | Repository `reviewThinkingLevel` is dormant and does not override the inherited selection |
| Repository review with no distinct review model at either scope | Resolved build model | Repository review Thinking Level, then Harness review Thinking Level, then resolved build Thinking Level |

The current Repository UI chooses a review Thinking Level source in a different order
from runtime resolution. This work must make the UI follow the runtime table rather than
encoding a second fallback policy.

## User Stories

1. As a Harness operator, I want a Thinking Level accepted only when the selected Agent Model advertises it, so that Settings cannot persist a known-invalid execution request.
2. As a Repository operator, I want build and review selections validated with their actual fallback semantics, so that inherited and explicit settings behave like Agent Turns.
3. As an operator changing a model in the browser, I want an incompatible draft Thinking Level cleared immediately, so that the new draft remains internally consistent.
4. As an operator opening Settings with a legacy incompatible Thinking Level, I want it shown as unavailable rather than silently deleted, so that I understand and control the correction.
5. As an operator correcting a legacy value, I want Save blocked until I choose a supported level or clear it, so that the invalid pair cannot be carried forward unnoticed.
6. As an operator whose selected model offers no Thinking Levels, I want a direct way to clear a stale level, so that the UI does not trap me in an unsaveable state.
7. As a direct GraphQL client, I want a structured error naming the invalid Thinking Level field, so that I can correct the exact input.
8. As an operator saving an unrelated setting, I want any carried-forward invalid pair identified before persistence, so that a whole-object write does not perpetuate it.
9. As an operator creating a Work Item from legacy settings, I want a known-invalid resolved pair rejected before work starts, so that no external model request is consumed.
10. As an operator with an existing Work Item after catalog drift, I want the next Agent Turn rejected before its Agent Backend process starts, so that the failure is deterministic and actionable.
11. As an operator using an explicit Agent Model with no explicit Thinking Level, I want the backend/model default retained, so that validation does not accidentally substitute the Harness level.
12. As an operator using a backend that reports no model catalog, I want the existing admission policy preserved, so that absence of catalog information is not treated as proof that every selection is invalid.
13. As a maintainer, I want one pure model/Thinking-Level validation policy reused at server boundaries, so that GraphQL and lifecycle admission cannot drift.
14. As a maintainer, I want persistence to remain independent of live Agent Backend inspection, so that database modules retain their current responsibility.

### 1. Model the catalog selection invariant once

Add or generalize a pure validator that receives a complete Agent Model catalog entry
and an optional Thinking Level. It should return a structured result that distinguishes:

- Agent Model absent from the current catalog.
- Thinking Level absent from the selected model's advertised `thinkingLevels`.
- Valid model with null Thinking Level.
- Valid model with an advertised Thinking Level.

Do not reduce the catalog to model IDs before validation. Preserve complete Agent Model
entries through the validation path.

The validator should compare canonical trimmed settings values, matching persistence
normalization. Empty or whitespace-only Thinking Levels are null. It must not mutate,
default, or substitute the input.

An Agent Model entry with `thinkingLevels: []` explicitly offers no Thinking Levels.
Null is valid for that model; every non-null level is invalid.

Keep catalog acquisition separate from the pure invariant. The existing settings path
may obtain a current Active Agent Backend status or use Agent Backend Preview for a
not-yet-Active backend. The pure validator should not know how the catalog was loaded.

### 2. Validate both settings mutations

Apply the invariant to both `updateConfig` and `updateRepositorySettings`.

For Harness Config:

- Validate the submitted build selection.
- Validate the effective review selection using the runtime fallback table.
- Preserve the rule that Harness Config requires a build Agent Model.

For Repository settings:

- Calculate the next Effective Agent Backend after applying the submitted override.
- Load the backend-scoped Harness preferences for that backend.
- Form the next Repository preference object using the mutation's existing whole-object
  semantics.
- Determine which submitted Thinking Levels contribute to the next resolved build and
  review selections using the runtime fallback table.
- Validate applicable explicit selections against the next Effective Agent Backend's
  catalog.
- Do not reject or delete dormant Repository Thinking Levels merely because a Harness
  selection currently takes precedence. The creation and pre-turn guards cover any
  future configuration change that makes a dormant value active.

Run validation inside the existing coordinated config-update section, before the
database write and Active Agent Backend synchronization. A failed validation must not
persist any field or change the Active Agent Backend set.

Continue the existing optimization where an unrelated Repository save with no explicit
catalog assertion and no applicable Repository Thinking Level does not require Preview
solely for validation.

### 3. Return existing structured settings errors

Use the existing invalid-config and invalid-Repository-settings error families. They
already allow these field names:

- `defaultThinkingLevel`
- `reviewThinkingLevel`

The message must name:

- The rejected Thinking Level.
- The governing Agent Model.
- The Agent Backend label.
- The model's currently advertised levels, or state that it offers none.
- The correction: choose an advertised level or clear the field to use the applicable
  fallback/backend default.

`reviewModel` is already checked for catalog membership. Preserve that behavior rather
than introducing a second review-model validation path.

### 4. Keep database persistence catalog-agnostic

Do not add Active Agent Backend, Preview, or catalog dependencies to the database
module. Database persistence remains responsible for trimming strings, converting blank
optional values to null, maintaining backend-scoped preference maps, and transactional
writes.

Do not add a database constraint or migration for dynamic model/Thinking-Level
membership. Catalogs vary by backend, provider mode, account, and runtime version, so a
static database constraint cannot express the invariant.

### 5. Make the browser correction path complete

Retain existing behavior that reconciles a draft Thinking Level when the operator
actively changes a model.

Add Thinking Level validity to both Harness Config and Repository Settings Save gates.
A loaded incompatible stored level remains visible as a synthetic unavailable option,
and Save remains blocked until the operator chooses a supported level or clears it.

When a model advertises no Thinking Levels, keep a correction control available. The UI
must allow a stale value to be changed to backend/model default; replacing the control
with a note while retaining an inaccessible stale value is insufficient.

Make Repository review Thinking Level options follow the runtime fallback table. In
particular, an inherited explicit Harness review model takes precedence over a
Repository build model.

Use scope-accurate copy:

- For an explicit Repository model, an empty Thinking Level means backend/model default.
- For an inherited complete Harness selection, describe the Harness selection.
- For review falling back to build, describe the build selection.

Opening Settings or refreshing a catalog must preserve an existing incompatible value
until the operator acts. Only an explicit model-change interaction may clear the browser
draft automatically.

### 6. Validate resolved selections at use time

Extend the existing Work Item catalog-admission policy to validate Thinking Levels as
well as model IDs.

At Work Item creation:

- Resolve the ordinary Work Item's complete build and review selections.
- Validate both roles against the current Ready catalog.
- Reject an invalid selection atomically and create nothing.

Before each Agent Turn:

- Validate the role actually used by that turn against the current Ready catalog.
- Fail before invoking the Agent Backend adapter when the model is absent or the
  Thinking Level is not advertised for that model.
- Include actionable guidance to correct Harness Config or Repository settings and
  retry/start again as appropriate.

Preserve the existing empty-catalog admission rule. A Ready Agent Backend that reports
no Agent Models provides no membership evidence, so ordinary Work Item admission may
continue to defer to runtime failure. This is different from a catalog that contains the
selected model with an empty `thinkingLevels` list; that is positive evidence that a
non-null Thinking Level is unsupported.

Use a Thinking-Level-specific Step Run reason rather than reporting the failure as
`agent_model_not_in_catalog`. If this adds a lifecycle reason code, edit the ontology
source and run lifecycle code generation; do not edit generated enums directly.

### 7. Keep documentation aligned

Update the Claude Code Agent Backend ADR's catalog-only decision to distinguish:

- Catalog-declared model/Thinking-Level mismatches, which fail at Settings or admission
  boundaries before process spawn.
- Compatibility failures that cannot be known from the advertised catalog, which may
  still surface as ordinary Agent Backend CLI failures.

Update GraphQL field documentation and operator-facing copy where needed so null-level
semantics match the runtime fallback table. If lifecycle reason vocabulary changes,
update the ontology and regenerate derived documentation/models through the established
ontology workflow.

## Acceptance Criteria

- [ ] Harness Config rejects an unsupported `defaultThinkingLevel` for its build Agent Model with a structured error naming `defaultThinkingLevel`.
- [ ] Harness Config rejects an unsupported `reviewThinkingLevel` for an explicit review Agent Model with a structured error naming `reviewThinkingLevel`.
- [ ] Harness Config validates review Thinking Level against the build Agent Model when review has no distinct model.
- [ ] Repository settings reject an unsupported Thinking Level that applies to an explicit Repository build or review Agent Model.
- [ ] Repository review validation follows Repository review, Harness review, then resolved build model precedence.
- [ ] Dormant Repository Thinking Levels are preserved rather than silently deleted.
- [ ] Null and whitespace-only Thinking Levels are accepted as unset and stored canonically.
- [ ] A model with no advertised Thinking Levels accepts null and rejects every non-null level.
- [ ] Validation works through both current Active status and Agent Backend Preview catalog paths.
- [ ] Validation failure performs no settings persistence or Active Agent Backend synchronization.
- [ ] Direct GraphQL callers cannot bypass the invariant.
- [ ] An incompatible browser draft is cleared when the operator actively changes models.
- [ ] A pre-existing incompatible level is preserved, visibly marked unavailable, and blocks Save.
- [ ] The operator can clear a stale level even when the selected model advertises no Thinking Levels.
- [ ] Repository review choices and labels reflect the runtime fallback order.
- [ ] Work Item creation rejects a legacy invalid resolved build or review pair and creates nothing.
- [ ] Pre-turn admission rejects a pair that became invalid after Work Item creation without invoking the Agent Backend adapter.
- [ ] A valid advertised level and a null level continue through creation and Agent Turns.
- [ ] An entirely empty Ready catalog retains the existing ordinary-work admission behavior.
- [ ] A catalog entry with an empty Thinking Level list rejects a non-null level.
- [ ] Existing catalog-only model validation remains unchanged.
- [ ] Relevant ADR, GraphQL, ontology/generated artifacts, and operator copy agree with the implemented semantics.

## Testing Decisions

Tests should assert externally visible behavior: accepted or rejected mutations,
structured error fields, persisted values, rendered correction controls, Work Item
creation, and whether an Agent Backend adapter was invoked. Avoid tests tied only to
helper names, React state names, source text, or SQL representation.

### GraphQL/application tests

Extend the existing catalog-validation tests for both settings mutations. Cover:

- Valid and invalid build Thinking Levels.
- Valid and invalid explicit review Thinking Levels.
- Review-with-build fallback.
- Repository review fallback through an explicit Harness review selection.
- Repository review fallback to resolved build.
- Active-status and Preview catalog acquisition.
- A model with no Thinking Levels.
- Null, empty, and whitespace normalization.
- Catalog unavailable while an applicable explicit selection is submitted.
- A whole-object unrelated save carrying an invalid applicable pair.
- No database update or Active-set synchronization after rejection.
- Correct structured field metadata and actionable message text.

### Browser tests

Extend the existing catalog-only Settings behavior tests. Cover both Harness Config and
Repository Settings:

- Changing models clears an incompatible draft level.
- A seeded incompatible stored level is visible as unavailable and blocks Save.
- Selecting a supported level permits Save and round-trips unchanged.
- Clearing uses the correct Harness/build/backend-default semantics for the scope.
- A stale value can be cleared when the model offers no levels.
- Repository review options use the same governing model as runtime resolution.
- Implicit form submission cannot bypass the Save gate.

### Lifecycle tests

Extend the existing no-spawn catalog-admission tests using controlled Agent Backend
adapters and real persistence where that is the established seam. Cover:

- Legacy invalid build selection at Work Item creation.
- Legacy invalid review selection at Work Item creation.
- Catalog drift before a later build or review Agent Turn.
- No adapter invocation on a known mismatch.
- Valid advertised and null levels still invoke the adapter.
- Entirely empty catalog behavior remains unchanged.

Update existing fixtures that pair a model with a level its fixture catalog does not
advertise; such fixtures should not accidentally depend on the bug this issue fixes.

### Verification

Run affected Nx targets for tests, type checking, linting, build, GraphQL generated
schema parity, ontology/lifecycle generated parity if applicable, and the relevant
Harness end-to-end target. Use the workspace package manager's Nx command rather than
invoking underlying tools directly.

## Implementation Sequence

1. Introduce the pure complete-catalog selection validator and make its unit tests cover every result variant. Completion criterion: model identity and Thinking Level can be validated without GraphQL, persistence, or an Agent Backend lookup.
2. Wire the validator into both coordinated settings mutations and add GraphQL behavior tests. Completion criterion: neither mutation can persist an applicable invalid pair, and all fallback cases have tests.
3. Complete the browser Save gates, correction controls, fallback source order, and copy. Completion criterion: every server rejection caused by pre-existing settings has an obvious browser correction path.
4. Extend creation and pre-turn admission, adding any lifecycle reason through ontology codegen. Completion criterion: legacy data and catalog drift fail before adapter invocation.
5. Update ADR/schema/domain documentation and run all affected verification. Completion criterion: generated parity checks and affected Nx targets pass.

## Out of Scope

- Automatically clearing, rewriting, or migrating persisted Thinking Levels.
- Changing Repository settings from whole-object model preference writes to a general
  patch mutation.
- Adding database constraints for dynamic catalog membership.
- Splitting preferences by provider mode.
- Inferring supported levels beyond what the Agent Backend catalog advertises.
- Probing a provider to prove a catalog-advertised combination can execute.
- Silently substituting another Thinking Level, Agent Model, or Agent Backend.
- Redesigning Explicit Work Item Execution Profiles; they should continue using their
  complete-profile validation and fail-closed behavior.

## Environment

- `ready-for-agent` v0.22.0
- OpenCode 1.18.16

## Further Notes

The implementation should preserve the architectural split established by issue #838:
catalog lookup and validation at application/API and process-spawn boundaries,
catalog-agnostic persistence, explicit operator correction for stale values, and
use-time protection for legacy data and catalog drift.

berenddeboer:
Settings could persist a valid Agent Model with a Thinking Level that
model does not advertise (for example `opencode/deepseek-v4-flash` with
`medium` when the catalog lists `low`, `high`, `max`). The browser
already derived levels from the catalog entry; `updateConfig` and
`updateRepositorySettings` reduced the catalog to model IDs and never
checked `thinkingLevels`. Whole-object Repository writes could carry an
invalid pair forward, and Work Item enrollment only checked model
identity.

This change treats advertised Thinking Levels as part of each catalog
entry's executable contract (#1073):

- Shared `validateCatalogSelection` distinguishes missing model, missing
  level, valid null, and valid advertised level. Whitespace is unset. A
  model with `thinkingLevels: []` accepts only null.
- Both settings mutations reject applicable invalid pairs with
  structured field errors before persist or Active-set sync. Dormant
  Repository levels are preserved. Validation uses Active status or
  Preview.
- Settings UI still clears an incompatible draft when the operator
  changes models. A stored incompatible level stays visible as
  unavailable and blocks Save, including when the model advertises no
  levels. Repository review options and empty-option copy follow runtime
  fallback (Repository review, then Harness review, then resolved
  build).
- Ordinary Work Item creation and pre-turn admission validate the
  resolved pair and fail before the Agent Backend adapter. Empty Ready
  catalogs still defer (no membership evidence). New Step Run reason:
  `thinking_level_not_in_catalog`.
- ADR 0047, GraphQL field docs, CONTEXT.md, and ontology stay aligned.
  Persistence stays catalog-agnostic.

Review follow-up: Settings now fail-closes applicable Thinking Levels
when the inherited governing model is absent from a Ready catalog
(including an empty Ready catalog). GraphQL
`Repository.reviewThinkingLevel` docs name the full null fallback chain.

Verification ran `work-item-lifecycle:test`, `graphql-api:test`,
`harness:test`, `lifecycle-model:test`, `db-service:test`, and typecheck
on the touched packages. Catalog-only Playwright (`@live-forge`) was
not run. Persisted invalid pairs are not auto-migrated; operators
correct them in Settings or they fail at creation/admission.

Closes #1073


berenddeboer:
Pushed `eef58c3dc7018d81545cf2f07e5077aeb3ba4039` to address the red **quality-gates** check.

**Cause:** `graphql-api:knip` failed on unused exports `applicableThinkingLevels` and `ThinkingLevelField` in `validate-agent-models.ts`. Both are only used inside that module.

**Change:** Stop exporting them. No behavior change.

**Verification:** `bunx nx run graphql-api:knip` and pre-commit `nx affected -t lint,typecheck,test` passed.

**Review feedback addressed:** none (no automated-review run or comment on this PR).

PR #1178, opened 2026-08-21
Add Azure DevOps as a third Forge

cgbarlow:
## Summary

Adds Azure DevOps as a third Forge, alongside GitHub and GitLab — by widening the existing closed union type rather than introducing a new Forge abstraction layer, matching this codebase's current convention (GitHub and GitLab are each an independent Effect-TS `Context.Service`, no shared interface today).

- New `packages/azure-devops-service` package mirroring `packages/gitlab-service`'s scaffold exactly (service interface, live implementation over raw `fetch`, hand-written test-double `Layer`, `bin/` CLI scripts).
- Remote detection for `dev.azure.com` / `*.visualstudio.com` clone URLs (`packages/local-git`), matched ahead of the existing GitLab catch-all.
- PAT-based auth (`AZURE_DEVOPS_EXT_PAT`), vault-first via Keymaxxer with permissive ambient fallback (matching GitLab's posture).
- Ready-issue listing via WIQL (`System.Tags CONTAINS 'ready-for-agent'`), native predecessor/successor blocking links surfaced as `blockedBy`.
- Full draft-PR lifecycle: create, update, mark ready for review, PR status checks (build validation / branch policy evaluations), merge, close-out with a completion summary.
- All 18 previously-implicit `if (forge === "gitlab") {...} else {/* github */}` call sites across the PR/merge state machine, auth/preflight, detection, and lifecycle/worker subsystems converted to explicit three-way branches.

## Verification

This was built and verified as a genuine multi-ticket epic on a fork (`cgbarlow/ready-for-agent`), then smoke-tested end-to-end against a real, independent Azure DevOps organization (not a fixture) — detect → authenticate → reconcile → build → open a draft PR → pass status checks → merge — landing a real merged PR against a real Azure DevOps Git repository. That live smoke test surfaced (and this PR fixes) several real-world gaps that unit tests alone hadn't caught:

- `_apis/connectionData` and `_apis/policy/evaluations` both require Azure DevOps' `-preview` API version flag on the otherwise-stable `7.1` version — omitting it fails every request with HTTP 400.
- `listReadyIssues` needs to filter to open-state work items itself; the WIQL tag query alone returns closed ones too, which would resurface historical work as fresh candidates on every reconciliation pass.
- Author-scoped relevance matching needs Azure DevOps' ADO-native `customDisplayName`, not the Entra/AAD directory `providerDisplayName` — real accounts can report different strings for each, and work items' `System.CreatedBy.displayName` uses the former.
- Azure DevOps' Git repository name can differ from its project name (a common real-world pattern — many orgs use one project, e.g. the ADO-created "Default" project, to host many differently-named repositories). PR/branch operations now resolve the repository name independently rather than assuming they match.

Existing GitHub and GitLab test suites pass unmodified; the new package's own suite (`bunx nx run azure-devops-service:test`) covers all service methods against the hand-written fake, matching `gitlab-service-test.ts`'s pattern. No new HTTP mocking library, no CI workflow changes needed (`nx affected` picks up the new package automatically via its `project.json` test target).

## Out of scope (by design)

- A shared `Forge` interface/abstraction layer — explicitly not introduced, to match the existing per-Forge `Context.Service` convention.
- Consolidating the four separate forge-union type declarations into one canonical type.
- Supporting more than one Azure DevOps org/project per repository (matches the existing one-remote-one-forge assumption already in `parseForgeRemote`).

Happy to split this into smaller PRs if that's easier to review — flagging up front that it's a large, multi-file change (a new ~2,000-line service package plus 18 converted call sites) since it's a genuinely new Forge, comparable in scope to the original GitLab support.


berenddeboer:
@cgbarlow best to have this as one PR, don't need it smaller.

PR #1197, opened 2026-08-23
fix(harness): soft-disable Keymaxxer when no vault exists

### LINKED ISSUE #1195: Soft-disable Keymaxxer when the operator never installed or initialized it
## Problem Statement

When an operator runs the local Harness (`bun run ready-for-agent start` / `nx run harness:dev`) on a checkout where they never installed or initialized Keymaxxer, startup still enables the Keymaxxer Sidecar. The Polling Auto-heal Job then fails in a loop:

```
ERROR (#25): Polling Auto-heal Job failed {
  jobId: "qjob-…",
  error: "Keymaxxer list failed",
  causeChain: [
    { name: "KeymaxxerError", message: "Keymaxxer list failed" }
  ]
}
```

Expected: missing / unusable Keymaxxer is ignored — the same product path as `KEYMAXXER_ENABLED=false` (ambient Forge auth, no Sidecar, no vault RPC, Auto-heal succeeds).

This is not a “binary completely absent from PATH” bug. That path already soft-disables. The failure is: the workspace `keymaxxer` npm package makes availability true, the Sidecar starts, `keymaxxer_list` fails (no vault), and Auto-heal fail-closes.

### Symptom path

1. Local Harness is `harness:dev` → `scripts/run-with-keymaxxer-sidecar.ts` → Vite `createApplication()`.
2. Availability is `isKeymaxxerAvailable(process.env)` (`packages/keymaxxer-service`): `KEYMAXXER_ENTRYPOINT` exists, or `Bun.which("keymaxxer")`.
3. Workspace `package.json` has `"keymaxxer": "0.2.2"` (devDependency). `bun install` creates `node_modules/.bin/keymaxxer`. Nx prepends that directory to PATH.
4. `isKeymaxxerAvailable` returns **true** even when the operator never installed Keymaxxer and has no vault.
5. Sidecar starts; `KEYMAXXER_SIDECAR_URL` is set. Application uses `sidecarKeymaxxerLayer` (`enabled` is unset, not `false`).
6. Startup `initialize` only waits for Sidecar TCP. It does **not** probe `keymaxxer_list`.
7. If the DB has any Repository, Polling Auto-heal calls `findSecret` → `keymaxxer_list`.
8. No vault: CLI `keymaxxer list` prints `no vault found. Run \`keymaxxer init\` first.`; MCP list returns `isError` → `KeymaxxerError("Keymaxxer list failed")`.
9. Auto-heal does **not** catch `KeymaxxerError` (unlike GraphQL `probeVaultSecret`). The job fails and retries with backoff.

`KEYMAXXER_ENABLED=false` works because `resolveKeymaxxerMode` / the wrapper set disabled env, `loadApplicationConfig` omits the Sidecar URL, and `disabledKeymaxxerLayer` sets `enabled: false`. Auto-heal then treats GitHub Repositories as ambient-credentialed and never calls list.

### Docker repro (this machine has Keymaxxer; the bug checkout does not)

Sandbox image `deboer-sandbox:gbrain-0.42.76.0-baseline` (node, no keymaxxer). Workspace mounted. Empty `HOME` (no vault). Host bun on PATH.

**Case 1 — workspace npm bin only (what `nx run harness:dev` sees after `bun install`):**

```
which=/…/worktree1/node_modules/.bin/keymaxxer
available=true
sidecar-up http://127.0.0.1:6057/…/mcp
[facade] waiting for vault unlock
RESULT: findSecret failed Keymaxxer list failed
```

That is the operator error.

**Case 2 — no `keymaxxer` on PATH at all:**

```
which=null
available=false
RESULT: unavailable → soft-disable
```

This already matches `KEYMAXXER_ENABLED=false`. The operator checkout is Case 1, not Case 2.

Direct CLI under the same Case 1 env: `keymaxxer list` → `keymaxxer: no vault found. Run \`keymaxxer init\` first.` (exit 1).

### Why existing tests miss it

| Check | What it covers | Why it misses this |
| --- | --- | --- |
| `resolveKeymaxxerMode({}, () => false)` | Injected “unavailable” | Never uses real `Bun.which` / workspace bin |
| `isKeymaxxerAvailable({}, () => false, () => false)` | Both probes stubbed false | Same |
| `harness:test` / `slow-test` / `smoke` / `e2e-no-backend` / `e2e-ui-history` | Force `KEYMAXXER_ENABLED=false` | Soft-disable is assumed, not earned |
| Auto-heal job-worker test | `KeymaxxerError` → postpone + retry | **Encodes fail-closed**, the opposite of product intent for “not installed” |
| GraphQL `probeVaultSecret` | `KeymaxxerError` → `{ kind: "unavailable" }` | Auto-heal does not share this |

Empty DB: Auto-heal never calls Keymaxxer (`listRepositories` is empty). The log appears only when at least one Repository exists (typical local `tmp/ready-for-agent.db`).

### Two cooperating defects

1. **Availability is “binary on PATH”**, not “operator Keymaxxer is usable”. Workspace `node_modules/.bin/keymaxxer` is a test/e2e tool, not an operator install. No vault still counts as available.
2. **Auto-heal is fail-closed on `KeymaxxerError`**. GraphQL credential activation already treats that as unavailable and falls back to ambient. Background Auto-heal should not ERROR-loop when the vault is missing.

Either fix alone is incomplete: (1) without (2) still ERROR-loops if the Sidecar is up and list later fails; (2) without (1) still starts a Sidecar and may prompt unlock on a machine that never wanted Keymaxxer.

### Intended product rule

Keymaxxer is optional. If it is not installed **or** not usable (no vault / list cannot succeed), behave as `KEYMAXXER_ENABLED=false`:

- no Sidecar
- `disabledKeymaxxerLayer` (`enabled: false`)
- ambient Forge auth
- Auto-heal succeeds (no `Keymaxxer list failed`)

## Solution

Soft-disable Keymaxxer whenever the operator has not actually made it usable, including the common monorepo case “`keymaxxer` resolves only to the workspace npm bin and there is no vault”. Auto-heal must not fail-close on `KeymaxxerError` from list/findSecret.

After the fix, Case 1 Docker (workspace bin, empty HOME, ≥1 Repository, `KEYMAXXER_ENABLED` unset) must be silent and healthy — same as explicit `KEYMAXXER_ENABLED=false`.

## User Stories

1. As an operator who never installed Keymaxxer, I want `bun run ready-for-agent start` to come up cleanly, so I can use ambient `gh` / `glab` without ERROR spam.
2. As an operator with Repositories already in the local DB, I want Polling Auto-heal to succeed without a vault, so Issue Polling still uses ambient credentials.
3. As an operator who only has the workspace `keymaxxer` npm package from `bun install`, I do not want that to count as “Keymaxxer installed”.
4. As an operator with no `~/.keymaxxer` vault, I want the Harness to skip Sidecar startup, so I am not prompted to unlock a vault I never created.
5. As an operator who sets `KEYMAXXER_ENABLED=false`, I want that path unchanged.
6. As an operator who installed Keymaxxer and has a vault, I want the Sidecar and vault-first credentials unchanged.
7. As an operator who sets `KEYMAXXER_ENTRYPOINT` to a real entrypoint and has a vault, I want that still to enable Keymaxxer.
8. As an operator who already has `KEYMAXXER_SIDECAR_URL` pointing at a live Sidecar, I want reuse unchanged (explicit existing-url mode).
9. As an operator on a machine where `keymaxxer` is not on PATH at all, I want the existing soft-disable to keep working.
10. As an operator watching Harness logs, I do not want repeating `Polling Auto-heal Job failed` / `Keymaxxer list failed` when Keymaxxer was never set up.
11. As an operator using GitHub Repositories with ambient `gh`, I want Auto-heal to treat those Repositories as credentialed when Keymaxxer is soft-disabled.
12. As an operator using GitLab Repositories with ambient `glab` / `GITLAB_TOKEN`, I want the same when Keymaxxer is soft-disabled.
13. As an operator using Azure DevOps with ambient PAT, I want the same (already ambient-only).
14. As a GraphQL caller, I want credential activation and Auto-heal to agree on “Keymaxxer unavailable → ambient”, so polling is not activated by one path and broken by the other.
15. As a contributor running `harness:dev` after `bun install`, I want a missing personal vault to be ignored, so I can work on UI/lifecycle without Keymaxxer setup.
16. As a contributor writing e2e, I want a suite that does **not** set `KEYMAXXER_ENABLED=false`, so this regression cannot hide behind the forced flag.
17. As CI on a runner that has the workspace `keymaxxer` package but no vault, I want ordinary (non-vault) jobs to stay vault-free without ERROR loops.
18. As an operator who later runs `keymaxxer init` and restarts, I want Keymaxxer to enable on the next start.
19. As an operator who uninstalls Keymaxxer and removes the vault, I want the next start to soft-disable again.
20. As an on-call reader of job logs, I want Auto-heal failures to mean a real repair problem, not “optional vault missing”.

## Implementation Decisions

- Change the Keymaxxer **mode / availability** decision used by both `run-with-keymaxxer-sidecar` (dev) and `resolveKeymaxxerMode` (production / operator binary). One rule, both entrypoints.
- “Available” must mean the operator can actually list secrets — not merely that a `keymaxxer` executable is on PATH. A workspace `node_modules/.bin/keymaxxer` with no vault is **not** available.
- Preferred signal: binary resolvable **and** a vault present (or an explicit Sidecar URL / working list). Do not treat “file exists on PATH” as sufficient. A probe that runs `keymaxxer list` is acceptable if it is fast, non-interactive, and cannot pop a passphrase dialog at boot.
- When unavailable, do exactly what `KEYMAXXER_ENABLED=false` already does: set that env, clear `KEYMAXXER_SIDECAR_URL`, use `disabledKeymaxxerLayer`.
- Defense in depth: Polling Auto-heal’s GitHub credential check must catch `KeymaxxerError` and treat it as unavailable / ambient — same idea as GraphQL `probeVaultSecret`. Do not postpone-and-ERROR-loop for “list failed” when Keymaxxer is optional.
- Do not start the Sidecar first and then discover there is no vault. Mode resolution happens **before** spawn (today’s shape is correct; the predicate is wrong).
- Do not require operators to export `KEYMAXXER_ENABLED=false` to use the local Harness.
- Do not change vault-backed live e2e (`harness:e2e` / `harness:e2e-live-forge`) or fixture-vault policy.
- Do not edit generated lifecycle / ontology files.
- Sidecar HTTP coming up is not proof Keymaxxer is usable; the facade listens before the keyholder/vault works.

## Testing Decisions

Test external behaviour: Harness with ≥1 Repository, `KEYMAXXER_ENABLED` unset, Keymaxxer not usable → no Sidecar-required path, no Auto-heal `Keymaxxer list failed`, GraphQL health true, ambient auth still works.

**Highest seam (required): live e2e.** Do **not** set `KEYMAXXER_ENABLED=false` (that is the current hole). Pattern after `@no-backend` PATH isolation:

- Strip ambient/system `keymaxxer` from the product PATH so the developer machine’s install cannot silently green the test.
- Leave (or prepend) a `keymaxxer` that exists but is not an operator vault — workspace `node_modules/.bin/keymaxxer` or a stub — so availability’s false positive is in play.
- Use a temporary `HOME` with **no** `~/.keymaxxer` vault.
- Seed at least one Repository (empty DB never calls list).
- Start the Harness the same way e2e already does (production supervisor is fine; it shares `resolveKeymaxxerMode` / `isKeymaxxerAvailable`).
- Assert: process stays up, GraphQL health is true, logs do not contain `Polling Auto-heal Job failed` / `Keymaxxer list failed`, Auto-heal acknowledges.

A Docker/sandbox Case 1 loop (workspace bin + empty HOME + `findSecret`) is a valid red-capable check and already reproduced `Keymaxxer list failed`.

**Supporting unit tests** (faster, not a substitute for the e2e):

- `isKeymaxxerAvailable` / `resolveKeymaxxerMode`: PATH has a keymaxxer bin, no vault → `disabled`.
- Same with no bin at all → still `disabled` (already true).
- Bin + vault (or explicit sidecar URL / `KEYMAXXER_ENABLED=false`) unchanged.
- Auto-heal: `findSecret` returns `KeymaxxerError` → job does not ERROR-loop; behaves like disabled/ambient. Today’s “retries until Keymaxxer succeeds” test must not keep requiring fail-closed for the optional-absent case.
- Prior art: `apps/harness/test/production-lifecycle.test.ts`, `packages/keymaxxer-service/test/mcp-layer.test.ts`, `apps/harness/test/job-worker.test.ts`, `packages/graphql-api` `probeVaultSecret`, `apps/harness/e2e/support/agent-backend-path.ts`.

## Out of Scope

- Making Keymaxxer required.
- Changing vault-backed e2e, fixture vault, or overnight/packed smoke that already set `KEYMAXXER_ENABLED=false`.
- Redesigning Sidecar capability URLs or MCP tools.
- Auto-heal backoff numbers except as needed so absence is not a failure.
- Teaching operators to export `KEYMAXXER_ENABLED=false` as the fix.

## Further Notes

- Local entry: `bun run ready-for-agent start` → `nx run harness:dev` → `run-with-keymaxxer-sidecar.ts`. Production / compiled binary uses `startProductionLifecycle` / `resolveKeymaxxerMode`. Fix both; they share `isKeymaxxerAvailable`.
- `application-config`: if `KEYMAXXER_ENABLED` is not `"false"`, `KEYMAXXER_SIDECAR_URL` is required. Soft-disable must set the flag **before** `createApplication`, as the wrapper already does when availability is false.
- Facade already knows `error: no vault found.` (`facade.test.ts`); it is not treated as “not installed” at mode resolution.
- Do not implement this ticket’s product change in the same session that only filed it.

berenddeboer:
Workspace `bun install` puts `keymaxxer` on PATH, so the Harness treated it as installed even with
no vault. Dev and production then started the Sidecar; Polling Auto-heal hit `Keymaxxer list failed`
and retried. Operators who never ran `keymaxxer init` should get the same path as
`KEYMAXXER_ENABLED=false`: ambient Forge auth, no Sidecar, no unlock prompt.

Availability now requires an on-disk vault (`vault.db` / `vault.meta.json`) under Keymaxxer's own
directory rules (`KEYMAXXER_DB_DIR`, then `$XDG_CONFIG_HOME/keymaxxer` only if that directory
exists, else `~/.keymaxxer`), plus a resolvable binary or `KEYMAXXER_ENTRYPOINT`. Mode resolution
still happens before Sidecar spawn. Explicit `KEYMAXXER_SIDECAR_URL` and `KEYMAXXER_ENABLED=false`
are unchanged. GitHub credential checks (GraphQL activation and Auto-heal) treat `KeymaxxerError` as
unavailable/ambient instead of fail-closing.

Verification: keymaxxer-service availability tests (including XDG vs HOME fallback), GraphQL GitHub
`KeymaxxerError` → ambient, job-worker Auto-heal no longer ERROR-loops, production-lifecycle mode
with a real PATH binary and empty HOME, and `harness:e2e-no-vault` (does not set
`KEYMAXXER_ENABLED=false`; stub `keymaxxer` on PATH, temp HOME, seeded Repository). Fork-safe CI
runs that suite.

Limitation: the live e2e infers Auto-heal success from the absence of fail-close logs (plus a
refresh side effect) rather than asserting queue acknowledge; the job-worker unit test covers
acknowledge.

Related: #1195.

Closes #1195

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -142,6 +142,10 @@ jobs:
       - name: Harness dev smoke
         run: bunx nx run harness:smoke
 
+      - name: Live Harness e2e (uninitialized Keymaxxer)
+        id: live-e2e-no-vault
+        run: bunx nx run harness:e2e-no-vault
+
       # Chromium once for both live suites.
       - name: Install Playwright Chromium
         working-directory: apps/harness
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -209,6 +209,10 @@ jobs:
       - name: Harness dev smoke
         run: bunx nx run harness:smoke
 
+      - name: Live Harness e2e (uninitialized Keymaxxer)
+        id: live-e2e-no-vault
+        run: bunx nx run harness:e2e-no-vault
+
       - name: Restore/install Playwright Chromium
         uses: ./.github/actions/playwright-chromium
         with:
--- a/apps/harness/project.json
+++ b/apps/harness/project.json
@@ -93,6 +93,16 @@
         }
       ]
     },
+    "e2e-no-vault": {
+      "executor": "nx:run-commands",
+      "options": {
+        "cwd": "{projectRoot}",
+        "command": "bun --conditions @ready-for-agent/source test test/keymaxxer-uninitialized-e2e.test.ts"
+      },
+      "inputs": ["default", "^production"],
+      "cache": true,
+      "dependsOn": ["^build"]
+    },
     "e2e-no-backend": {
       "executor": "nx:run-commands",
       "options": {
--- a/apps/harness/project.json
+++ b/apps/harness/project.json
@@ -97,7 +97,7 @@
       "executor": "nx:run-commands",
       "options": {
         "cwd": "{projectRoot}",
-        "command": "bun --conditions @ready-for-agent/source test test/keymaxxer-uninitialized-e2e.test.ts"
+        "command": "bun --conditions=@ready-for-agent/source test test/keymaxxer-uninitialized-e2e.test.ts"
       },
       "inputs": ["default", "^production"],
       "cache": true,
--- a/apps/harness/scripts/run-unit-tests.sh
+++ b/apps/harness/scripts/run-unit-tests.sh
@@ -23,9 +23,12 @@ effect_ignore=(
   "**/application-runtime-disposal.test.ts"
   "**/production-sse-idle-timeout.test.ts"
 )
-# Also exclude the slow SSE suite from the default bun path (historical).
+# Also exclude the slow SSE suite from the default bun path (historical)
+# and the Keymaxxer uninitialized live e2e (must not inherit this runner's
+# KEYMAXXER_ENABLED=false; run via harness:e2e-no-vault).
 bun_ignore=(
   "**/production-sse-idle-timeout.test.ts"
+  "**/keymaxxer-uninitialized-e2e.test.ts"
   "${effect_ignore[@]}"
 )
 
--- a/apps/harness/src/server/job-worker.ts
+++ b/apps/harness/src/server/job-worker.ts
@@ -29,6 +29,7 @@ import {
   POLLING_AUTO_HEAL_BACKOFF,
   POLLING_AUTO_HEAL_KEY,
   enqueuePollingAutoHealJob,
+  githubRepositoryHasCredential,
   issuePollFailureBackoff,
   repairPollingSchedules,
   sampleIssuePollingDelay,
@@ -128,11 +129,7 @@ const repositoryHasGitHubCredential = Effect.fn(
 )(function* (projectPath: string) {
   const keymaxxer = yield* KeymaxxerService
   if (keymaxxer.enabled === false) return true
-  const credential = yield* keymaxxer.findSecret({
-    provider: "github",
-    account: projectPath,
-  })
-  return credential !== null
+  return yield* githubRepositoryHasCredential(keymaxxer, projectPath)
 })
 
 const repositoryHasCredential = Effect.fn("JobWorker.repositoryHasCredential")(
--- a/apps/harness/test/job-worker.test.ts
+++ b/apps/harness/test/job-worker.test.ts
@@ -2984,136 +2984,132 @@ describe("Job worker", () => {
       }),
   )
 
-  it.live("auto-heal retries with backoff until Keymaxxer succeeds", () =>
-    Effect.gen(function* () {
-      const database = DbServiceLive.pipe(Layer.provideMerge(DatabaseTest))
-      const queue = SqliteQueueServiceLive.pipe(Layer.provideMerge(database))
-      let findSecretCalls = 0
-      const keymaxxer = Layer.succeed(KeymaxxerService, {
-        initialize: Effect.void,
-        findSecret: ({ account, provider }) =>
-          Effect.gen(function* () {
-            findSecretCalls += 1
-            if (findSecretCalls < 3) {
+  it.live(
+    "auto-heal treats KeymaxxerError as ambient and does not ERROR-loop",
+    () =>
+      Effect.gen(function* () {
+        const database = DbServiceLive.pipe(Layer.provideMerge(DatabaseTest))
+        const queue = SqliteQueueServiceLive.pipe(Layer.provideMerge(database))
+        let findSecretCalls = 0
+        const keymaxxer = Layer.succeed(KeymaxxerService, {
+          initialize: Effect.void,
+          findSecret: () =>
+            Effect.gen(function* () {
+              findSecretCalls += 1
               return yield* new KeymaxxerError({
                 operation: "findSecret",
-                message: "Keymaxxer unavailable",
+                message: "Keymaxxer list failed",
               })
-            }
-            return provider === "github" &&
-              account === `${repository.projectPath}`
-              ? `GITHUB_TOKEN`
-              : null
-          }),
-        findSecrets: () => Effect.die("not used"),
-        hasSecret: () => Effect.die("not used"),
-        addSecret: () => Effect.die("not used"),
-        runWithSecrets: () => Effect.die("not used"),
-      })
-      const lifecycle = Layer.succeed(WorkItemLifecycle, {
-        maxDurations: {
-          create_worktree: Duration.minutes(5),
-          install_dependencies: Duration.minutes(15),
-          implement: Duration.hours(2),
-          assess_changes: Duration.minutes(5),
-          pre_commit: Duration.hours(2),
-          review: Duration.hours(1),
-          commit: Duration.minutes(5),
-          create_pr: Duration.minutes(10),
-          watch_pr_status_checks: Duration.minutes(5),
-          resolve_pr_merge_conflict: Duration.hours(2),
-          investigate_pr_status_checks: Duration.hours(2),
-          mark_pr_ready_for_review: Duration.minutes(5),
-          decide_pr_merge: Duration.minutes(15),
-          merge_pr: Duration.minutes(5),
-          close_issue: Duration.minutes(5),
-          local_cleanup: Duration.minutes(5),
-        },
-        implementNow: unused,
-        implementWith: unused,
-        implementLocally: unused,
-        implementAllWithAutoMerge: unused,
-        queue: unused,
-        recoverOrphanedStepRuns: Effect.succeed(0),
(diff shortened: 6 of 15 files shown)


berenddeboer:
Pushed `d8881c2` to this branch.

**What changed:** `harness:e2e-no-vault` now invokes `bun --conditions=@ready-for-agent/source test …`. Bun 1.4 treats the spaced form `bun --conditions VALUE test` as `bun run test`, which exited immediately with `Script not found "test"` (186ms, no suite output). Topology now asserts the equals-form command.

**Verification:** reproduced the CI failure locally with the spaced command; the equals form runs `test/keymaxxer-uninitialized-e2e.test.ts` and `test/topology.test.ts` (10 pass). Pre-commit `nx affected` lint/typecheck/test passed.

**Review feedback addressed:** none (Claude Code Review was skipped; no review comments).

PR #1273, opened 2026-09-08
feat: make Settings, Save, and Agent Turns share one Agent Model catalog (#1272)

### LINKED ISSUE #1272: Refresh Agent Model catalogs on Settings open and share freshness with Save and Agent Turns
## Problem Statement

Repository Settings can offer an Agent Model that Save rejects because the dropdown and server validation use different catalog snapshots.

Reproduced against a running harness version 0.31.0:

1. The harness default Agent Backend was Codex Build; the Repository had an OpenCode override that was already Active.
2. OpenCode configuration was updated to include `azure/gpt-6-astra` while the harness was running.
3. Opening Repository Settings offered Astra in the Build model dropdown.
4. Selecting Astra with High Thinking Level and saving failed with: `Agent Model "azure/gpt-6-astra" is not in the current OpenCode Agent Model catalog. Choose a model the Agent Backend currently offers.`
5. The saved model remained unchanged.

Live API inspection confirmed that OpenCode Agent Backend Preview returned a fresh catalog containing Astra, while Active OpenCode status still held an older catalog without Astra. Preview did not update Active status. The dropdown used Preview because OpenCode differed from the harness default, while Save preferred Active status.

This is an inconsistent catalog-authority/freshness problem, not a database persistence failure or invalid model ID. Operators should not need to know about Recheck or restart the harness to select a newly configured model.

## Solution

Refresh the relevant Agent Model catalog when Harness Settings or Repository Settings opens. For an already-Active Agent Backend, publish the successful refreshed catalog to the running harness so that the dropdown, Save validation, and subsequent Agent Turn validation use the same authoritative snapshot.

Keep Save fast by reusing a sufficiently recent server-held catalog. If the catalog has become stale while Settings was open, refresh it before validating and persisting the selection. Do not unconditionally launch another catalog inspection on every Save.

Keep Agent Backend Preview non-activating for an inactive, unsaved backend selection. Browsing Settings must not change the selected backend or activate an otherwise inactive backend.

## User Stories

1. As an operator, I want newly configured models to appear when I open Settings, so that I do not need to restart the harness.
2. As an operator, I want a model offered by the dropdown to pass catalog validation when it remains available, so that Settings does not contradict itself.
3. As an operator, I want Save to reuse the catalog just fetched for Settings, so that a routine save stays fast.
4. As an operator, I want Save to refresh an expired catalog before validation, so that leaving Settings open does not silently preserve obsolete availability information.
5. As an operator, I want the next Agent Turn to recognize a newly saved model, so that fixing Save does not simply move the same rejection into execution.
6. As an operator, I want the same behavior for the harness default and already-Active Repository overrides, so that backend selection does not determine whether Settings works.
7. As an operator, I want build and review models and their Thinking Levels validated against the refreshed catalog, so that supported model/effort combinations remain executable.
8. As an operator, I want removed models to remain visible as unavailable rather than being replaced, so that my selections are never silently changed.
9. As an operator, I want browsing another backend to remain a Preview, so that opening Settings does not change Agent Turn routing.
10. As an operator, I want a slow or failed required refresh to show a bounded, actionable error without losing my form edits, so that I can retry safely.
11. As an operator, I want catalog refreshes not to interrupt running Agent Turns or alter explicit execution-profile selections, so that ongoing work is preserved.
12. As an operator, I want overlapping Settings requests to reuse an in-flight refresh, so that multiple views do not cause redundant catalog discovery.

## Implementation Decisions

- Use the existing Agent Backend inspection/catalog boundary and Active backend registry rather than adding a second independent cache in Settings.
- Refresh on opening Harness Settings and Repository Settings for the effective backend. Switching the backend selection should load the correct catalog without violating Preview versus activation semantics.
- For Active backends, a successful refresh must update the authoritative server-held catalog consumed by Settings, Save validation, and Agent Turn admission. Merely accepting a client-provided catalog or refreshing only the dropdown is insufficient.
- Track freshness per backend on the server. Define and document a short freshness window; a successful Settings-open refresh should normally cover the subsequent Save. Use the same freshness policy for Harness and Repository saves.
- A required stale-on-Save refresh must complete before model/Thinking Level validation and persistence. Save must not claim fresh validation after that refresh fails.
- Reuse in-flight refreshes for the same backend and preserve protection against older inspection results overwriting newer results.
- Retain existing catalog-only selection, inheritance, clearing, unavailable-model preservation, and backend availability rules. Do not silently substitute models or Thinking Levels.
- Refreshing an Active backend catalog must not restart the harness, cancel running Agent Turns, or rewrite Work Item execution profiles. Previewing an inactive backend must not activate it; activation remains subject to the existing Save gates.
- Provide bounded loading/error behavior for inspection. Keep unsaved form values after errors and avoid treating a loading catalog as a definitive empty catalog.
- No database workaround or new persisted model catalog is needed by this request. Update relevant documentation if freshness behavior changes the documented Preview/Recheck contract.

## Testing Decisions

Prefer behavioral tests at existing boundaries, not source-text assertions or a new parallel test harness.

- Extend the existing catalog-only Agent Model browser acceptance coverage with the exact failure pattern: start the harness with an older catalog, add a model to the controlled backend while the harness remains running, open Repository Settings for an already-Active non-default backend, select it, save successfully, and reopen to confirm persistence. Do not restart the harness between changing the catalog and opening Settings.
- Use existing GraphQL integration coverage to prove that Settings refresh and Save agree, a fresh Save does not repeat discovery, and an expired catalog refreshes before validation/persistence. Cover both Harness Settings and Repository Settings.
- Use a controlled clock and inspection results in the existing Active backend tests to cover the freshness boundary, deduplicated overlapping refreshes, failed/slow inspection, out-of-order results, and a backend leaving the Active set during refresh. Avoid wall-clock sleep tests.
- Verify that newly added models are accepted, removed models or Thinking Levels are rejected without silent substitution, and failed validation does not persist settings or activate a previewed backend.
- Cover the connection to Agent Turn validation: after the successful refresh and save, the next turn must not reject the newly added model using the previous catalog. Preserve existing captured-backend and explicit-profile behavior; do not call a real paid model in tests.
- Preserve existing Preview success/failure tests proving that inspecting an inactive backend leaves the Active set unchanged.

Existing prior art includes catalog-only Agent Model browser scenarios, GraphQL settings/model validation integration tests, Active backend preview/recheck and concurrency tests, and captured-backend/Implement With catalog-drift admission tests. Existing UI refresh-on-open source assertions alone do not prove server catalog freshness.

## Out of Scope

- Periodic background catalog polling or automatic refresh on every Agent Turn.
- Unconditionally refreshing on every Save when a recent authoritative catalog is available.
- Removing catalog validation, adding free-text model selection, or automatically substituting a different model.
- Changing model inheritance, backend-switch idle gates, Work Item routing, or explicit execution profiles.
- Optimizing OpenCode startup or guaranteeing a fixed discovery latency across all Agent Backends.

## Further Notes

Measured locally during diagnosis: `opencode models --verbose` took about 1.665 seconds; fresh catalog discovery through the harness took approximately 1.75-1.79 seconds. These are observations, not a universal latency guarantee. They support paying discovery cost on Settings open and reusing the result on a prompt Save, rather than imposing the same delay twice.

A targeted Recheck OpenCode is the current workaround because it replaces the Active catalog; the requested behavior should make that manual step unnecessary for normal Settings use.

berenddeboer:
Repository Settings could offer an Agent Model that Save rejected. Opening Settings used a Preview catalog, while Save validated against a stale Active catalog. Operators had to Recheck or restart the harness to pick a newly configured model.

Opening Harness or Repository Settings now always Previews the effective backend. When that backend is already Active, a successful Preview publishes the catalog for Save and later Agent Turns for one minute. Save reuses that snapshot instead of inspecting on every persist, and inspects again only after the window expires. Preview of an inactive, unsaved backend still does not activate it. A failed Preview keeps the previous Ready catalog and readiness.

Recheck and Settings Preview can overlap. Recheck stays concurrent with Recheck. Recheck waits only for an in-flight Settings Preview before giving up, so a failed Preview cannot drop a Recheck that already finished inspecting. A successful Settings Preview of an Active backend returns the published catalog when Recheck won, so the dropdown matches Save.

Verified with Active backend tests (freshness window, overlapping refreshes, Recheck vs Preview races, drop during refresh), GraphQL Settings/Save agreement for Harness (reuse, expiry, failed refresh) and Repository reuse, and the catalog-only end-to-end scenario that adds a model while the harness is running and saves it from Repository Settings without a restart.

Out of scope: periodic catalog polling, inspecting on every Save when a fresh snapshot exists, and extra Agent Turn-after-save or Repository expiry GraphQL coverage (same server path as the Harness tests).

Closes #1272


berenddeboer:
Pushed `6ccf8135c44ad576247d4cb470c6e2fb69e40d9a` to fix the red live-Forge and UI-history Settings checks.

**What changed.** A failed Settings Preview was still treated as an in-flight catalog load, so the dropdown stayed on “Loading catalog…”. Harness Settings also hid the whole form behind “Loading settings...” while Preview ran. Failed Preview is now a settled error, and catalog loading stays on the model fields. Save remains blocked until Preview succeeds.

**Verification.** `bun test` on the Harness Settings and Repository Settings source tests; commit hook ran affected lint, typecheck, and test.

**Review feedback.** Claude Code Review was skipped (no reviewer steps, no review comment). No automated-review feedback to address or decline.

The earlier quality-gates failure was a transient OpenCode download (`curl: (35) Connection reset by peer`), not this catalog change. This push retriggers the PR workflow.

berenddeboer:
Pushed `316b16b3a9942e53b3da4019a3dad9448be37dd3` to fix the remaining UI-history failure.

**What changed.** Browser Forward opened Settings, started a catalog Preview, then session prepare discarded that fetch. “Loading catalog…” never cleared, so the Back/Forward Settings scenario timed out. Preview now re-runs after prepare on routed open (Forward / direct). The `harness` job was red only because UI-history failed; live-Forge was already green.

**Verification.** Harness Settings and Repository Settings source tests; commit hook ran affected lint, typecheck, and test.

**Review feedback.** Claude Code Review was skipped (no reviewer steps, no review comment). None declined.

PR #1283, opened 2026-09-11
fix: keep the page still while a native modal is open

### LINKED ISSUE #1279: Prevent background scrolling while a modal dialog is open
## Problem

When a dialog is open, mouse-wheel scrolling can still move the page behind it. Only scrollable content inside the active dialog should respond; the background must remain stationary until the dialog closes.

## Investigation

Static inspection found a shared fix opportunity without replacing the dialog system:

- Dialogs use native `dialog` / `showModal()` and the shared `dialogPanel` recipe in `apps/harness/src/ui.ts:414-421`. The recipe has no explicit scroll containment or background lock. Body recipes at lines 440-448 provide layout, not a separate scrolling container.
- `apps/harness/src/styles.css:245-257` has no document scroll lock; lines 844-849 only style the backdrop.
- The background uses normal document scrolling. Existing Session Telemetry browser tests assert preservation of `window.scrollY`, but do not exercise mouse-wheel input while a dialog is open.
- Applicable declarations under `apps/harness/src/`: Harness Settings (`routes/__root.tsx`), Repository settings and Repository-not-found (`home-page-content.tsx`), Session usage (`session-usage-dialog.tsx`), and the shared Implement With wrapper (`implement-with-dialog.tsx`), used for both leaf and parent issues, including loading states.
- Session usage has a nested scrollable Agent Turn Tail list that must remain usable.

This is static evidence, not a runtime-confirmed diagnosis. Opening the running app at localhost:3000 redirected to login, so browser reproduction and validation remain part of implementation. No application code was changed. Searches of open issues for scroll and dialog found no duplicate.

## Implementation Direction

Implement a shared modal-aware document scroll lock plus containment of scrollable dialog content. Confirm the actual document scrolling element and computed dialog overflow in a browser first.

A narrowly scoped CSS rule using `html:has(dialog:modal)` and `overflow: hidden` is a candidate, subject to supported-browser validation. Add appropriate overscroll containment to the shared panel recipe and make viewport sizing/vertical overflow explicit only if needed. Containment alone is insufficient: wheel input over the backdrop or a short dialog must also be blocked from scrolling the background.

- Follow actual modal state rather than route state: first-run Harness Settings and Implement With can be local-only.
- Release the lock on close or removal, but retain it while any modal remains open. Avoid independent per-dialog style mutations and cleanup races.
- Preserve originating scroll position, focus behavior, and Back/Forward/direct-entry behavior. Avoid scrollbar-induced layout jumps.
- Do not introduce a new dialog library or unrelated navigation refactor.

## Acceptance Criteria

- [ ] With a genuinely overflowing background at a nonzero scroll offset, opening any modal prevents background movement from mouse-wheel or trackpad scrolling, including over the backdrop and a short/non-overflowing dialog.
- [ ] Long dialog content still scrolls normally. Repeated outward scrolling at its top or bottom does not chain into the background.
- [ ] Cover Harness Settings, Repository settings, Repository-not-found, Session usage, and leaf/parent Implement With, including local-only opening and loading-to-form transitions.
- [ ] Nested scrollable content, including Agent Turn Tail, remains usable. Long forms and final actions remain reachable on desktop and narrow/short viewports; touch scrolling must not leak to the background on mobile.
- [ ] Closing through each applicable route (Close, Cancel, Escape, successful Save, browser Back, or unmount) restores normal background scrolling without losing the prior position. Forward/reopening locks again.
- [ ] Failed Save or a pending-save dismissal guard keeps the background locked while the dialog remains open. If multiple modals are present, closing one does not unlock the background prematurely.
- [ ] Existing navigation/focus behavior is unchanged, with no noticeable horizontal layout jump or unwanted overflow.
- [ ] Add real-browser regression coverage proving the background can scroll before opening and after closing, remains stationary while open, and the dialog itself can scroll, including boundary and backdrop cases. Use actual wheel input with pointer placement and eventual position assertions, not just programmatic `scrollTop` updates or DOM-only tests.

## Verification Guidance

Extend the seeded, vault-free UI-history browser suite. Existing features under `apps/harness/e2e/features/` are `settings-browser-history.feature`, `repository-settings-browser-history.feature`, and `session-telemetry-browser-history.feature`. The latter already tests originating scroll-position preservation. Use representative long/short dialog tests plus coverage that all variants participate in the shared lock; avoid an exhaustive cross-product.

Follow ADR 0048 (browser-addressable dialog routes), ADR 0021 (live Harness E2E / UI-history scope), and the shared native-dialog design system. Do not submit Implement With actions that start real Work Items merely to test scrolling. Run relevant checks through Nx.

berenddeboer:
Wheel and trackpad input could still move the Harness behind an open dialog. Only the active dialog (and nested lists such as Agent Turn Tail) should scroll until it closes.

The lock follows native modal state (`html:has(dialog:modal)`), not the route, so first-run Settings and Implement With stay locked even when they never get a URL. Overscroll is contained on the shared panel and on `dialog:modal`. Implement With keeps one dialog mounted from preference loading through the form so the lock is not dropped on that swap. Failed Save and pending Save (including Back blocked while Saving…) keep the page locked.

UI-history browser coverage uses real wheel input with pointer placement: overflowing background before open and after close, backdrop and short-dialog cases, long Settings including boundary overscroll, nested Agent Turn Tail, first-run, leaf and parent Implement With without submitting work, Failed Save, pending Save, Forward/reopen, and a narrow viewport. Layout width is asserted in Chromium.

Limitations: this is CSS-only, so iOS pan-behind-modal is not separately proven. Classic scrollbars may still shift layout slightly; always-on gutter would inset the full-bleed chrome. Close-path coverage is representative (Cancel, Close, Back), not every Escape/Save/unmount combination.

Closes #1279

### DIFF
--- a/apps/harness/e2e/features/modal-dialog-scroll-lock.feature
+++ b/apps/harness/e2e/features/modal-dialog-scroll-lock.feature
@@ -1,0 +1,178 @@
+@ui-history
+Feature: Lock document scrolling while a modal dialog is open
+  Mouse-wheel and trackpad input must not move the page behind an open
+  native modal. Dialog content and nested lists still scroll. Closing
+  restores the originating offset. Coverage is representative rather than
+  an exhaustive close-path cross-product (issue #1279).
+
+  Scenario: Harness Settings locks an overflowing background and restores it
+    Given the Harness has Session Telemetry fixtures
+    And the Harness has a configured default build model
+    When I open Completed page 2 directly
+    And I cancel the Harness settings dialog if present
+    And I prepare an overflowing page background at a nonzero scroll offset
+    Then mouse-wheel input can move the page background
+    When I open Harness settings from the masthead
+    Then the Harness settings dialog is visible
+    And the page background scroll position is unchanged
+    And the page layout width is unchanged
+    When I wheel over the dialog backdrop
+    Then the page background scroll position is unchanged
+    When I wheel over the open dialog
+    Then the open dialog can scroll
+    And the page background scroll position is unchanged
+    When I wheel past the open dialog scroll boundary
+    Then the page background scroll position is unchanged
+    When I cancel the Harness settings dialog
+    Then the Harness settings dialog is hidden
+    And the page background scroll position is unchanged
+    And mouse-wheel input can move the page background
+    When I open Harness settings from the masthead
+    And I go back in the browser
+    Then the Harness settings dialog is hidden
+    And the page background scroll position is unchanged
+    When I go forward in the browser
+    Then the Harness settings dialog is visible
+    When I wheel over the dialog backdrop
+    Then the page background scroll position is unchanged
+
+  Scenario: Failed Save keeps the background locked
+    Given the Harness has Session Telemetry fixtures
+    And the Harness has a configured default build model
+    When I open Completed page 2 directly
+    And I cancel the Harness settings dialog if present
+    And I prepare an overflowing page background at a nonzero scroll offset
+    And I open Harness settings from the masthead
+    And Harness settings Save is forced to fail
+    And I save Harness settings expecting failure
+    Then the Harness settings dialog is visible
+    And a settings save error is shown
+    When I wheel over the dialog backdrop
+    Then the page background scroll position is unchanged
+
+  Scenario: Pending Save keeps the background locked
+    Given the Harness has Session Telemetry fixtures
+    And the Harness has a configured default build model
+    When I open Completed page 2 directly
+    And I cancel the Harness settings dialog if present
+    And I prepare an overflowing page background at a nonzero scroll offset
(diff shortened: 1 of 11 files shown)


berenddeboer:
Pushed `308e30c31931720265820bbc093fc1c50afeaa15` to address CI.

**Changes**
- Stop exporting `ImplementWithModalDialog` (knip unused export after loading moved into the same dialog). `bun nx run harness:knip` now passes.
- Restore the recorded page offset under the modal lock before asserting it holds, so routed Settings opens that adjust `window.scrollY` do not fail the overflowing-background scenario.

**Verification:** `harness:knip`; pre-commit lint/typecheck/test.

**Review feedback:** none declined — Claude Code Review was skipped and there were no review comments.

PR #1320, opened 2026-09-19
fix: contain agent and repository subprocesses per invocation

### LINKED ISSUE #1318: Contain agents and repository commands; terminate every owned process on timeout or Interrupt
## What to build

Encapsulate every Harness-launched Agent Turn and native repository command in an invocation-owned execution boundary established **at launch**. When the Harness concludes that the invocation timed out or was interrupted, the agent/command and every local process it started must be terminated before a replacement attempt can run against the same Work Item. Retrying must not accumulate abandoned tools, hook, Nx, build, subagent, or server processes.

Apply one ownership contract across OpenCode, Codex Build, Claude Code, Grok (including ACP continuation), and native repository-command execution including pre-commit validation and Commit hooks. Account for wrappers that start descendants in separate process groups/sessions, double-fork, exit, and reparent **before cleanup begins**. Killing only the immediate process, its original group, or a PPID snapshot at shutdown is insufficient. Keep ownership scoped to the invocation, never broad process-name matching.

On Linux, investigate an invocation-specific cgroup/systemd unit or an equivalent durable kernel-backed ownership mechanism. A POSIX process group is useful for graceful signalling but does not by itself enforce this guarantee. Do not replace the global build lock with a global execution lock: independent Work Items must remain independent.

## Incident evidence

Process Focus Work Item `wi-01M2SB9V73K14XAXNV6DKMD1GY`, Issue processfocus/monorepo#3058, Codex Session `01a0b30c-e186-7eb3-ba3c-949e05a6ce5a`:

- Two Commit Step Runs timed out after 30 minutes each; another subsequently timed out too.
- Old affected-static verification processes continued running after their recorded attempts ended and were reparented to user systemd.
- Twelve abandoned verification roots across Process Focus #3058, #3060, and #3079 were waiting on a leaked frontend-build lock. Recovery terminated 79 processes belonging to those twelve trees, then the two orphaned auth/GraphQL servers.
- Explicit interruption of the two active Pre-commit Step Runs also left descendant verification processes alive, requiring targeted OS-level cleanup.

The Process Focus build lock is being fixed independently in https://github.com/processfocus/monorepo/issues/3112. The Harness defect remains: an ended Step Run can leave processes alive that interfere with later attempts.

## Verified agent-backend gap

The existing shared CLI runner already starts agents as POSIX process-group leaders and performs bounded SIGTERM → SIGKILL cleanup. OpenCode, Codex Build, Claude Code, and Grok's CLI path use this runner; Grok's ACP continuation launches detached and explicitly installs the same tree-kill finalizer.

However, the helper enumerates descendants by PPID **when cleanup starts**, not durably from launch. A descendant which has both escaped the group and been reparented before that point is lost.

Verified against the real shared runner with a fake agent and real OS processes, without model/API calls:

1. Agent starts a short-lived tool wrapper.
2. Wrapper starts `setsid sleep 100` with standard streams redirected and records its PID, then exits.
3. Agent remains alive until a 500 ms timeout, or an Effect fiber interruption after 500 ms.
4. The Harness kills the agent. The worker remains alive under user systemd, in its own process group/session.

Both timeout and interruption probes failed the assertion that the worker must be dead. The test finally block explicitly killed its own survivor. Existing shared-runner process-tree/setsid/finalize tests passed (3 tests), and Grok ACP timeout/interruption tests passed (2 tests); those tests do not exercise reparenting before cleanup.

This is a separately reproduced agent-ownership defect, not a claim that the exact original production agent exited by this mechanism.

## Acceptance criteria

- [ ] Agent Turns and native repository commands have explicit invocation-scoped execution ownership established before their code can spawn children. It survives intermediate wrapper exit/reparenting and covers descendants even when they create new process groups/sessions before termination begins.
- [ ] The ownership mechanism is wired through every supported agent backend, including Grok ACP continuation, and through native hooks/install/build commands. Starting a new turn for the same durable Session gets a new invocation boundary.
- [ ] Timeout and operator Interrupt trigger bounded graceful shutdown followed by forceful termination of remaining owned processes. Cleanup is awaited before a replacement attempt executes in the same worktree.
- [ ] An ended attempt cannot silently leave owned descendants alive. Verify that the owned execution boundary is empty; cleanup failure is surfaced with actionable diagnostics rather than reported as successful cleanup. Preserve timeout/interruption as the initiating reason while reporting any cleanup failure separately.
- [ ] Normal agent exit/final output also releases invocation-owned subprocesses; shutdown must not depend on the original agent PID remaining alive. Harness shutdown/restart ownership and recovery are explicit, so abandoned invocation boundaries can be identified and reaped safely.
- [ ] Worktree contents, staged changes, Session identity, and Step Run history remain intact through interruption and retry.
- [ ] Processes belonging to another Work Item, the Harness, an interactive operator session, or unrelated applications are not killed.
- [ ] A subprocess integration fixture reproduces an agent/tool-wrapper/worker tree where a worker performs setsid/double-fork and its intermediary exits **before** timeout or Interrupt. Both paths must kill the worker and leave an unrelated concurrent invocation alive. Verify successful agent exit with a surviving child as well.
- [ ] A native hook/wrapper/build/server fixture retains a file lock. After timeout and after operator Interrupt, the owned processes are gone and retry can acquire the resource immediately. Exercise the production process-ownership seam rather than a mocked kill function.
- [ ] Include a termination-resistant child and demonstrate bounded escalation, plus a concurrent unrelated process that survives.
- [ ] Document and test platform-specific ownership and termination behaviour. Do not silently advertise complete containment when a fallback is only best-effort PPID/group scanning; inability to establish required containment must have an explicit supported policy and actionable outcome. Follow existing domain/ontology workflow if new Step Run reasons are needed.

## Scope

Reliable local subprocess ownership for all Harness-started agents and repository commands. Do not solve this by increasing Step Run timeouts, skipping hooks, killing all Bun/Node processes, or adding another shutdown-only PPID scan. An agent protocol cancellation acknowledgment is not proof of OS-process cleanup. This issue is independently implementable and is not blocked by the Process Focus build isolation/performance tickets.

berenddeboer:
Agent and hook descendants could escape process groups, outlive timed-out or interrupted attempts, and retain locks that blocked retries.

Give every agent turn and native repository command a fresh Linux cgroup managed by a transient systemd user service. Ownership survives setsid, double-forking, and reparenting. Cleanup performs bounded TERM-to-KILL escalation, verifies emptiness, and completes before retry. Normal exit and final output also release descendants; watchdogs and recovery handle Harness crashes. Cleanup failures block worktree reuse and preserve timeout or interruption as the initiating outcome.

Cover all agent backends, including Grok ACP continuation, plus repository commands and credential-sidecar pushes. Preserve missing-executable guidance and permission diagnostics. Adapt CI execution to the required delegated user hierarchy.

Verification: real subprocess tests cover escaped, termination-resistant workers, concurrent invocation isolation, crash recovery, cleanup failure, and Commit hooks holding locks. Retry tests verify immediate lock acquisition and preserved staged changes. Regression tests cover executable errors and PATH fallback. The full staged pre-commit hook passed after correcting a discovery test's startup allowance; the final review reported no findings.

Compatibility: local execution now requires Linux cgroup v2 with cgroup.kill (Linux 5.14+) and a delegated systemd user manager (systemd 254+). Unsupported hosts explicitly refuse execution. This is process lifecycle containment, not a security sandbox.

Closes #1318


berenddeboer:
Pushed 8b643c1f443edef1f948abdc461cec1d6a3010dc to fix the failing quality gate.

The spawn-error fixtures depended on Claude being installed: executable preflight returned a missing-CLI error on CI before the injected permission error could run. They now use the running test executable, and the permission test asserts that the exact injected error survives. Production behavior is unchanged.

Verification: reproduced the failure with Claude absent from PATH, then confirmed the same test passes after the fix. All 173 agent-backend tests pass with that restricted PATH; lint, typecheck, Knip, and the full commit hook also passed.

Review feedback: no automated-review feedback was present. The review job was skipped with no executed steps or review comment. No feedback was declined or deferred.

berenddeboer:
Pushed 4c61c4c1e6cf7cc907e747388dcc596b1068acae to fix the failing browser-history check and its aggregate harness check.

The CI trace showed a settings Save batched with background queries. The failure fixture returned a single GraphQL error object for the entire batch, so the dialog displayed a response-length error instead of the simulated save failure. The fixture now preserves response order and count, forwards unrelated queries, and prevents the failed mutation from reaching the server. Application behavior is unchanged.

Verification: a deterministic mixed-batch regression failed before the fix and passed afterward. It also checks that the stored settings remain unchanged. All 73 browser-history scenarios pass, including the original failed-Save scenario; Harness lint, typecheck, Knip, and the normal commit hooks also passed.

Review feedback: the latest automated-review job was skipped with no executed steps or review output. No review feedback was present, declined, or deferred.

